PR #25, opened 2018-02-26
chore(ffe-formatters): Fix non-deterministic tests

### LINKED ISSUE #23: Tests for formatPercentage in ffe-formatters fail under certain conditions
Two test cases of ffe-formatter fail on certain yet to be discovered conditions on two machines in my possession (Linux + Mac). It seems Number.toLocaleString prefixes negative numbers with either an hyphen minus (char code 45) or a minus sign (char code 8722), whereas the later causes the test to break.

https://github.com/SpareBank1/designsystem/blob/209c8f5d7d7e6d1d19d308b779fe35b75338016c/packages/ffe-formatters/src/formatPercentage.spec.js#L17

https://github.com/SpareBank1/designsystem/blob/3e3cff94ade0f398dc24f22c0df75a9d34d62c69/packages/ffe-formatters/src/formatPercentage.spec.js#L33-L35

https://github.com/SpareBank1/designsystem/blob/209c8f5d7d7e6d1d19d308b779fe35b75338016c/packages/ffe-formatters/src/formatPercentage.js#L12-L15

kwltrs:
On some environments the tests for formatPercentage with negative numbers fail because the minus sign is represented by different characters (char code 45 or 8722).

Fixes #23

### DIFF
--- a/packages/ffe-formatters/src/formatPercentage.spec.js
+++ b/packages/ffe-formatters/src/formatPercentage.spec.js
@@ -14,7 +14,6 @@ describe('format percentage', () => {
 
     test('formats numbers as percentages', () => {
         expect(formatPercentage(30)).toBe(`30${NON_BREAKING_SPACE}%`);
-        expect(formatPercentage(-73.2)).toBe(`-73,2${NON_BREAKING_SPACE}%`);
         expect(formatPercentage(125)).toBe(`125${NON_BREAKING_SPACE}%`);
     });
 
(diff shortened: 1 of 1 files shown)


wkillerud:
I ran Node v9.6.1 on my machine with en_US locale, and the test passes just fine. It also passes on CI. I'm unsure we want to add this workaround if it's only happening on your machines. It seems like fixing a symptom rather than a root cause, but I understand it's frustrating.

kwltrs:
It happens on a company issued machine and another system, currently important for our build pipeline. Everything was fine up to now and I predict this will pop up somewhere again.

And this is a really hard to spot issue. The test runner output is: `Expected value to be: "-98,765 %" Received: "−98,765 %"`. This change will save anybody running into this message in the future a lot of confusion and time.

Further, I disagree that this a symptom of a problem. Somewhere inside node's implementation of Number.toLocalString some component thinks it should use this unicode character instead of one from the lower ascii table. Even though I'm not a typography expert, my gut feeling tells me a character called "minus sign"—and `&minus;` as html entity—is a as valid prefix as "hyphen-minus". It might even be the better option. So this change is not a work around around a bug.

wkillerud (approved):

henrikhermansen (approved):
I agree we should make sure to handle this. JavaScript can be surprising and unpredictable in this way. When two different systems utilize two different characters as a minus sign, it seems natural our tests should run fine in both systems.

selbekk (approved):
Looks good! Another approach could be to standardize even further, and change the minus sign in the formatting function itself. 

That would simplify testing, as well as make sure the output is the same in any browser or environment.

Issue #4, opened 2018-02-26
Expand button "X" animates in from the top-right corner


wkillerud:
When the expand button animates in its "close X" the X comes down from and is animated out to the top right. This animation looks odd and should perhaps be redesigned. Extract the attached ZIP to see a video recording of the issue. 

[Knapp.zip](https://github.com/SpareBank1/designsystem/files/1758543/Knapp.zip)

Relevant code can be found here: https://github.com/SpareBank1/designsystem/blob/master/packages/ffe-buttons/less/expand-button.less

selbekk:
This will be fixed by the pending button rewrite PR

wkillerud:
Fixed by #27

PR #27, opened 2018-02-27
Reimplement buttons

### LINKED ISSUE #12: ffe-buttons-react: TertiaryButton only supports rightIcon if there is also a leftIcon
`TertiaryButton` doesn't support an icon on the right side of the label unless there is also an icon on the left side. This is an unintended effect of the `simpleContent` prop in `Button` being set to `true` (which renders a simpler markup without support for icons.) unless there is a `leftIcon`.

Quickfix: set `simpleContent` to `false` if there is either a `leftIcon` or a `rightIcon` ([here](https://github.com/SpareBank1/designsystem/blob/e7b68cd44af010bd9e7debc621137f3a6fe84874/packages/ffe-buttons-react/src/TertiaryButton.js#L19))

Perhaps better: get rid of the whole `simpleContent` concept and restructure the components internally to support the different use cases in a simpler way

selbekk:
This pull request re-implements both `ffe-buttons` and `ffe-buttons-react` in a long-awaited major breaking change.

Please give me feedback on the way I've solved things, as I want to try to find the best possible API before merging it.

Fixes #12 and #4


packages/ffe-buttons-react/src/ExpandButton.js
```diff
@@ -1,37 +1,49 @@
...
-            })}
-            onClick={onClick}
+        <Button 
+            aria-expanded={isExpanded}
+            buttonType="expand" 
+            className={classNames(
+                { 'ffe-button--expanded': isExpanded },
+                className,
+            )}
+            leftIcon={isExpanded && <KryssIkon />}
```
wkillerud: If I provide a `leftIcon` I would expect it to apply only to the label in its unexpanded state. Now it will override the `KryssIkon` when `isExpanded` is `true`. `KryssIcon` should perhaps instead be rendered instead of `children`?

That being said, `ExpandButton` should probably not build on `BaseButton` here in React IMO. We don't have an `isLoading` state, and so rendering the spinner and its wrapper seems a bit pointless.

wkillerud: `InlineBaseButton` would be a better parent I think :) Only, this isn't necessarily inline. Hmm..

packages/ffe-buttons/less/base-button.less
```diff
@@ -0,0 +1,245 @@
...
+    &:hover {
+        box-shadow: 0 1px 2px 1px rgba(0, 0, 0, 0.2);
+    }
+
+    &:active {
+        box-shadow: inset 0 1px 1px 0 rgba(0, 0, 0, 0.25);
+    }
+
+    &--action&--ghost {
+        &,
```
wkillerud, line 35: Unintended, or should the default and focus state look the same?

wkillerud (changes requested):
I'd love the commit messages even more if they had a "before" example as well - how does the markup look today and what are the key changes? (no modifiers inside the button, no label wrapper, but a wrapper for the spinner instead).

I'd also love it if the commit message for the breaking change in React-land mentions the most important breaking changes instead of pointing to the .md immediately.

packages/ffe-buttons/less/base-button.less
selbekk: According to the [original implementation](https://github.com/SpareBank1/designsystem/blob/master/packages/ffe-buttons/less/action-button.less#L16-L21), this is how it should look.

packages/ffe-buttons-react/src/ExpandButton.js
selbekk: Some good points! 
The reason why this abstraction was used is because the `ffe-button--expand` extends the look and feel of the `ffe-button--secondary`. It shares the same common less-class as well (`ffe-button`), and it follows a lot of the same usage patterns.

As you might have noticed, I added more full-featured prop types to each of the components than they had previously. This is to specify the supported API better than what the previous implementation did. I considered destructuring waay unsupported properties so that misuse would be even harder, but decided against it. I believe undocumented usage should be discouraged, not defensively coded against.

That being said, I'll try to implement a custom component that tries to solve this a bit better.

selbekk:
Okay, so I did a few changes:

- Improved the commit logs for easier migration
- Simplified the spinner implementation, now it only needs a single DOM element
- Renamed prop `buttonRef` to `innerRef` (which is a bit easier to standardize across the codebase)
- Re-added `ariaLoadingMessage` as a documented prop, but with a sensible default.
- A few documentation improvements

wkillerud (approved):

henrikhermansen (approved):
Ser veldig bra ut :1st_place_medal:

selbekk:
I would really love a review from one of the core maintainers (@kwltrs @antidecaf) on this before it's merged. The changes in the LESS API is big and breaking, so I would love to have most people on board 👯‍♂️

antidecaf (approved):
LGTM!

selbekk:
@kwltrs or @antidecaf  do you mind merging this whenever you're ready to release it? Since it's a complete rewrite, I think it's beneficial if we try to release it as soon as possible after a merge to master.

kwltrs (approved):

PR #28, opened 2018-02-27
chore: Use suggested tool for running prettier

### LINKED ISSUE #26: Consider using pretty-quick for running prettier
Prettier tends to create mangled diffs every once in a while. While researching how to fix this, I noticed that the updated [prettier website](https://prettier.io/) suggests using [`pretty-quick`](https://github.com/azz/pretty-quick) instead.

Should we change out `precise-commit` for `pretty-quick` in order to improve the commit experience? It does not support patch-wise formatting, but I think that's what's ruining prettier's output in the first place.

selbekk:
This commit replaces `precise-commits` with `pretty-quick`. Apart from
the obviously more clever name, `pretty-quick` has a few benefits to it:

- It's the suggested tool of Prettier, making it more heavily used and
  more maticiously maintained.
- It has really nicely formatted output
- It doesn't support patch-based running.

Now, this last part is the real reason why this change is warranted.
When you run `prettier` on a patch (and that patch alone), it doesn't
take into consideration the code around it. This led to mangled code
(that no longer even compiled), and post-commit unstaged changes. Nah.

Instead, `pretty-quick` formats any file that's been touched for a given
commit. Sure, it's going to format more code, leading to larger diffs,
but I'm confident the upsides outweight its downsides.

Fixes #26


kwltrs:
So far the prettier-commit-hook has caused some pain. Therefor I suggest we remove it as a hook and make it a regular npm run script, e.g `npm run make-staged-files-prettier`. As soon we feel the prettier setup is stable and fits our needs, it can make its comeback as commit hook.

selbekk:
Sure that's possible, but it's a bit sketchy to only run the prettier formatter whenever you want to. You can always skip the commit hook with `-n` - wouldn't it be safer (and provide more feedback, quicker) to keep it as a commit hook, and iterate on the setup until it fits our needs instead?

kwltrs:
I feel prettier should iterate as regular run script as I believe a commit-hook should be more reliable than the two prettier hooks we tried so far.

selbekk:
Very well, I'll amend the pull request

package.json
```diff
@@ -47,7 +47,6 @@
         "eslint": "^4.13.1",
         "eslint-config-prettier": "^2.9.0",
         "handlebars": "^4.0.11",
-        "husky": "^0.14.3",
```
wkillerud: Do we not use Husky for commitlint?

https://github.com/marionebl/commitlint#getting-started
https://github.com/typicode/husky/blob/master/HOOKS.md#hooks

wkillerud (dismissed):

package.json
selbekk: giphy 21
Will fix!

kwltrs: Good catch. Yes, we need husky for commitlint.
https://github.com/SpareBank1/designsystem/blob/5a01f4abf7a5e88c214d7bc7ccacdafe889e0c88/package.json#L14

PR #35, opened 2018-02-28
fix(ffe-datepicker-react): Press Enter to close calendar

### LINKED ISSUE #14: ffe-datepicker-react: pressing Enter while focusing the input field with a valid date should close the calendar
If I edit a date in the input field with my keyboard and hit Enter I expect the calendar to close so I can tab to the next input field, given the date is valid.

henrikhermansen:
This commit makes a simple improvement to keyboard navigation of the
datepicker by allowing the calendar to be closed by pressing Enter while
the date input field has focus.

This solution is slightly different from the suggested solution. While the
suggestions solution said the calendar would only close if the date is
valid, this solution closes the calendar regardless and triggers validation.
This allows us to show an error message to the user, rather than keeping
them stuck in the calendar.

This commit also removes the `onKeyPress` listener from Input.
The reason for doing so is that the dual key listeners conflicted with the
initial changes of this commit.
I did extensive testing to assure this is a safe change to make. I am also
reassured by the fact that we don't use this listener anywhere else in our
code, while we use the `onKeyDown` listener extensively.

This fixes #14

### DIFF
--- a/packages/ffe-datepicker-react/src/datepicker/Datepicker.spec.js
+++ b/packages/ffe-datepicker-react/src/datepicker/Datepicker.spec.js
@@ -6,6 +6,7 @@ import Calendar from '../calendar';
 import Datepicker from './Datepicker';
 import i18n from '../i18n/i18n';
 import ErrorTypes from '../datelogic/error-types';
+import KeyCode from '../util/keyCode';
 
 const defaultProps = {
     value: '',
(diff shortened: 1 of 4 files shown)


wkillerud:
LGTM. Could you add a unit test that verifies this behavior? :woman_scientist:

wkillerud (dismissed):

selbekk (dismissed):
Great job bud!

selbekk (dismissed):

henrikhermansen:
@wkillerud I added a commit with a couple of new unit tests, before squashing it all to a single commit. See `Datepicker.spec.js`.

packages/ffe-datepicker-react/src/datepicker/Datepicker.spec.js
```diff
@@ -6,6 +6,7 @@ import Calendar from '../calendar';
 import Datepicker from './Datepicker';
 import i18n from '../i18n/i18n';
 import ErrorTypes from '../datelogic/error-types';
+import { KeyCodes } from '../../../ffe-account-selector-react/src/util/types';
```
wkillerud: Cheeky ;) ~~I prefer duplication over reuse in this case. Either that or move `types` to a shared location somehow, but that seems out of scope.~~

There's already one here in [ffe-datepicker-react](https://github.com/SpareBank1/designsystem/blob/f0db1201b8cccab09967ec2aae3782a1d2ae3add/packages/ffe-datepicker-react/src/util/keyCode.js#L3)

wkillerud (changes requested):

packages/ffe-datepicker-react/src/datepicker/Datepicker.spec.js
henrikhermansen: Oh my.. I intended to use the same constants as defined within the package, but as I wrote "keyCodes" instead of "keyCode" IntelliJ was being overly helpful in importing this for me :)
Fixed in the most recent commit, which is an amend commit of the previous squashed commits.

selbekk (approved):

wkillerud (approved):

Issue #33, opened 2018-02-28
Consider removing all `peerDependencies`


wkillerud:
#### Which package is this issue related to?

:globe_with_meridians: All the things

#### Describe your issue

When using packages with a React and Less version we have a `peerDependency` on the Less version from the React version. A lot of Less packages also have a `peerDependency` on `@sb1/ffe-core` (even some React packages, like accordion do this). [These tend to get out of sync](https://github.com/SpareBank1/designsystem/search?utf8=%E2%9C%93&q=%22peerDependencies%22&type=), as it seems Lerna ignores them. This again creates a whole set of warnings on `npm install` that _this and that peer dependency is missing_. Some of these peerDependency requirements may also conflict for absolutely no reason.

The peerDependencies were originally added to give some kind of hint that you need this other package for the package to work as intended. For React and Less pairs I feel this is a given, and the peerDependencies just add noise.

What I'm suggesting is that we remove it all.

`@sb1/ffe-core` is a bit tricky. Maybe now that we use Lerna we could instead make `@sb1/ffe-core` a _direct_ dependency? This way we:

- Can remove `peerDependencies` (:+1:)
- A breaking change in `core` also triggers a major release of all dependencies, regardless of feature use (:+1: / :-1: )

henrikhermansen:
How do we keep track of version requirements between packages?
Say you add a new modifier to `ffe-buttons`.
When you add a property to utilize this modifier in `ffe-buttons-react` the consumers need to know that you need minimum version x.y.z of `ffe-buttons` to actually use the new prop in `ffe-buttons-react`.

Without peerDependencies I guess you need to manually check change logs.
Is this good enough?
Are there other solutions?

wkillerud:
A valid concern. Communicating that is difficult. 

Maybe it's OK to expect consumers to upgrade both packages (and check the logs of both) when wanting to use a new feature?

Lerna (for valid reasons) [delegates to us to maintain peerDependencies](https://github.com/lerna/lerna/issues/1018).  May be it's just one of those things we have to deal with...

selbekk:
IMO there are three things we need to decide on:

1. If a React component depends on a different local React component package - should it be listed as a dependency, or as a peer dependency + dev dependency?
2. If a React component applies classes that's provided from a different local LESS package - should it be listed as a peer dependency, or not at all?
3. If a LESS package depends on a different LESS package (i.e. ffe-core), should it be listed in dependencies, or as a peer dependency + dev dependency?

[...]

PR #36, opened 2018-02-28
fix: Correct peerDependencies after breaking change

wkillerud:
The peerDependencies were not updated after the version bump when we
added scopes to the packages, so users get spammed with peer dependency
warnings. This version fixes the peerDependencies so installing the
package should no longer produce a warning that another sb1 package is
missing (granted it's actually there).

### DIFF
--- a/packages/ffe-accordion-react/package.json
+++ b/packages/ffe-accordion-react/package.json
@@ -35,8 +35,7 @@
     "jest": "^22.4.0"
   },
   "peerDependencies": {
-    "@sb1/ffe-accordion": "^3.0.10",
-    "@sb1/ffe-core": "^11.0.2",
+    "@sb1/ffe-accordion": "^4.0.0",
     "react": "^15.0.0 || ^16.0.0"
   },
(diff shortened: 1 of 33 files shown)


selbekk (approved):
As far as I can tell, we haven't really standardized what should be a peer dependency and what shouldn't.

Should we create a different issue to go through all dev and peer dependencies, and apply a common standard to them all?

wkillerud:
@selbekk I've proposed getting rid of them alltogether in #33. We can use that issue to discuss other options? :)

selbekk:
Sounds great @wkillerud 👍

ivarni:
I've always been under the impression that if a package depends on a different package being installed then it should be a peerDependency. If it depends on a *specific* version of that package then that should be indicated as loosely as possible.

This is how they were meant to be used when we started making `-react` packages. I am pretty sure it was discussed at the weekly meetings because I remember there being complaints from those who were at npm v2 at the time because they couldn't just ignore peer deps.

PR #42, opened 2018-03-01
ffe-account-selector-react: Disregard formatting when searching for account numbers

### LINKED ISSUE #6: ffe-account-selector:  search shows no results after entering the fifth number
When searching for an account it can seem like the search requires we include the separator to get a match on an account number. We should not force users to do this. Ignore separators when searching.

See attached images.
sok1
sok2

wkillerud:
Fixes #6 

These changes make sure to apply the same string replace to both the query and the queried. This way we can be sure that formatting doesn't factor into the search.

screen shot 2018-03-01 at 18 30 47


selbekk (approved):
Great job !  LGTM!

PR #43, opened 2018-03-01
ffe-account-selector-react: Honor the propType contract for selectedAccounts

wkillerud:
Issue #5 was opened requesting we allow `selectedAccounts` to be `undefined`. This PR adds an empty array to `defaultProps` and updates the documentation to reproduce this issue locally and get a working live example.


selbekk (approved):
Nice! 👍

PR #50, opened 2018-03-02
Rewrite ffe-radio-button-react

selbekk:
This commit rewrites the `ffe-radio-button-react` implementation to improve, simplify and modernize the API.

This is an initial pull request - it's not complete, but complete enough to get some feedback.

A quick summary of this PR is the following:

There are now four components:
- `<RadioButton />` - a simple radio button choice
- `<RadioSwitch />` - a binary radio button choice
- `<RadioButtonGroup />` - a wrapping component simplifying the use of the previous two
- `<RadioBlock />` <-- this guy is new - well, for React, at least.

There's another internal component called `<BaseRadioButton />` that is the base unit creating the three first components.

Note that the `<RadioButtonGroup />` is requiring a [render-prop (or function-as-a-child)](https://reactjs.org/docs/render-props.html) - this is something that's not heavily used in FFE yet. It _is_ used by `ffe-form-react`'s `<InputGroup />` component, and it works just fine. It's a different, less magical way of intuitively passing props to children.

Let me know what you think!

Fixes #50


packages/ffe-radio-button-react/src/RadioButton.md
```diff
@@ -0,0 +1,88 @@
...
+            </RadioButton>
+            <RadioButton {...inputProps} value="none">
+                Ikke kunde enda
+            </RadioButton>
+        </React.Fragment>
+    )}
+</RadioButtonGroup>
+```
+
+Noen valg kan være mer kompliserte, og ikke like selvforklarende. Da kan man
```
wkillerud: Is this really by design, or is the tooltip there for historical reasons? Seems like an odd place to explain complex options - I'd think a tooltip explaining the whole RadioButtonGroup would be better.

packages/ffe-radio-button-react/src/RadioBlock.js
```diff
@@ -0,0 +1,64 @@
...
+            selectedValue,
+            title,
+            value,
+            ...inputProps
+        } = this.props;
+
+        const isSelected = checked || selectedValue === value;
+
+        return (
+            <div className="ffe-radio-block">
```
wkillerud: Is a `className` prop relevant here?

packages/ffe-radio-button-react/src/RadioButtonGroup.js
```diff
@@ -0,0 +1,80 @@
...
+            >
+                {description && 
+                    <legend className="ffe-form-label">
+                        {description}
+                        {tooltip &&
+                            <Tooltip {...tooltipProps}>{tooltip}</Tooltip>
+                        }
+                    </legend>
+                }
+                {children(buttonProps)}
```
wkillerud: I approve this pattern :rocket: So much more flexible for wrappers like this.

packages/ffe-form/less/radio-switch.less
```diff
@@ -28,6 +28,10 @@
     @media screen and (min-width: @breakpoint-sm) {
         margin-bottom: 0;
     }
+
+    & + input + .ffe-radio-switch {
```
wkillerud, line 32: This change is non-breaking I take it (from the commit message)?

wkillerud (commented):
Liking the changes to the `RadioButtonGroup`, and it'll be nice to have a `Group`-less `RadioSwitch` :+1:

packages/ffe-form/less/radio-switch.less
selbekk: Yep that's right. Let me explain: 

In the previous implementation, two radio switch elements (there are typically two of them) had to be separated by inline styles. That was the way this was solved in the existing React implementation at least. This approach solves this issue by a clever CSS selector. 

To make it a bit easier to understand what I'm doing - here's the typical DOM structure of a pair of radio switches:

```html
<input id="yes" type="radio" class="ffe-radio-input" />
<label for="yes" class="ffe-radio-switch">Yes</label>

<input id="no" type="radio" class="ffe-radio-input" />
<label for="no class="ffe-radio-switch">No</label>
```

I guess I could have used the class instead, but in reality that doesn't matter much.

packages/ffe-radio-button-react/src/RadioButtonGroup.js
selbekk: I mentioned this in a [thread on Twitter](https://twitter.com/selbekk/status/969960538003427328) to Kent Dodds - the guy that popularized this render prop pattern. He said

> If the state you want to share is implicit then compound components are great. Otherwise render props are much more powerful and flexible.

I think we want render props for this. The [react docs even have their own section pm them](https://reactjs.org/docs/render-props.html) if you want to learn more.

packages/ffe-radio-button-react/src/RadioBlock.js
selbekk: Nice catch! Yes, yes it is. I'll add it tonight.

selbekk: Yep! Nice catch. We might need that for some of the child elements as well, but I think I'll leave that part until we actually DO need them.

packages/ffe-radio-button-react/src/RadioButton.md
selbekk: Probably historical. I guess we can ask the designers! 

@adidick - should each radio button choice be able to have its own tooltip? With today's implementation, it looks a bit silly.

packages/ffe-radio-button-react/src/RadioButtonGroup.js
```diff
@@ -0,0 +1,80 @@
...
+            inline,
+            name,
+            onChange,
+            selectedValue,
+        };
+
+        return (
+            <fieldset 
+                className="ffe-fieldset" 
+                onChange={this.onChange}
```
wkillerud: This should be `onChange` (sans `this`), right? To use the prop. With that this can also be a stateless functional component.

packages/ffe-radio-button-react/src/RadioButtonGroup.js
wkillerud: For consistency could we also have the option for single child use the "old fashioned way" without render prop, [like InputGroup](https://github.com/SpareBank1/designsystem/blob/d6baf4bb982c1bc1e2d98b72f1dac315a4a5d68e/packages/ffe-form-react/src/InputGroup.js#L27)?

wkillerud (dismissed):

packages/ffe-radio-button-react/src/RadioSwitch.js
```diff
@@ -0,0 +1,42 @@
...
+        leftValue,
+        rightLabel,
+        rightValue,
+        ...rest
+    } = props;
+
+    return (
+        <Fragment>
+            <BaseRadioButton
+                className="ffe-radio-switch"
```
wkillerud: We need the same `classNames` handling here as in `RadioButton`, otherwise a `className` set on the component will override the FFE class.

packages/ffe-radio-button-react/src/RadioButtonGroup.js
```diff
@@ -0,0 +1,80 @@
...
+            'aria-invalid': ariaInvalid, 
+            inline,
+            name,
+            onChange,
+            selectedValue,
+        };
+
+        return (
+            <fieldset 
+                className="ffe-fieldset"
```
wkillerud: I'd love a `className` prop here as well :place_of_worship:

packages/ffe-radio-button-react/src/RadioButtonGroup.js
selbekk: I'm actually thinking no. In my opinion, allowing for both ways was wasn't a good solution. It doubles the API, which makes it hard to understand (and even harder to explain) when to use what. Sure it's less boilerplate, but it's also more of a "hm what should I do here" and "aww crap I have to switch from one to the other" moment prone technique.

If there are very strong opinions on this I'll budge, but I kind of think this is the way to go.

packages/ffe-radio-button-react/src/RadioButtonGroup.js
selbekk: Very true - this was some left-behind garbage from a previous iteration. Nicely spotted!

packages/ffe-radio-button-react/src/RadioSwitch.js
selbekk: I agree!

packages/ffe-radio-button-react/src/RadioButtonGroup.js
selbekk: No problem mister!

selbekk:
@wkillerud is killing it (wink wink) with the review comments here - anyone else interested in this? cc @SpareBank1/designsystem  @SpareBank1/designsystem-collaborators

packages/ffe-radio-button-react/src/RadioButton.md
adidick-zz: I'm not sure exactly what you mean here @selbekk. Can we take a look at it together when I'm back (currently home with sick kids)?

selbekk: @adidick This is what we're talking about - per-radio button tooltips:

image

Should this be allowed at all, or should we just have them per group - like this?

image

Hope your kids feel better soon!

packages/ffe-radio-button-react/src/RadioButtonGroup.js
wkillerud: > It doubles the API, which makes it hard to understand (and even harder to explain) when to use what.

One child - compose as you normally would. Two or more children - use the render prop. I think that's simple enough to explain. Coupled with a helpful error message when trying to compose as normal with two or more children, I don't see the problem as a user. It's worked very well for `InputGroup` in my opinion.

> it's also more of a "hm what should I do here"

Having the thing break when I try to compose as normal with a single child is also a bit of a "hm what should I do here" :smile: 

> it's also more of a (...) "aww crap I have to switch from one to the other" moment prone technique

True, and it's a bit annoying - I'll give you that. Still, I prefer reducing the amount of boilerplate for the `RadioSwitch` case over always using the render prop. It's not a complex rewrite going from one pattern to the other.

adidick-zz:
It’s hard to give a concrete answer without knowing if there are any cases where we might need one info button on each radio button. My gut tells me that that’s overkill and that we should be doing more to simplify the possible answers by presenting all the info up front so that when it comes to choosing the user can do it easily with a full understanding of the question.

packages/ffe-radio-button-react/src/RadioButtonGroup.js
wkillerud: @selbekk and I discussed this and he has me convinced that render prop only is the way to go. It avoids the case where a consumer just wraps the two or more children in a `Fragment` or whatever to make the error go away. Less ways to mess things up :+1:

selbekk: I love these discussions! :tada: 

Reality is, there was a lot of confusion when I introduced this new technique for input groups. People did what people do and basically tried to hack around the new API by providing a wrapping DIV (which then received the `aria-invalid` and `id` props, breaking the intended effect of having this component in the first place. :disappointed: 

I agree that it's... nice to have, if you know the API or read the docs extensively, but I think an API with fewer options is better than a "simple-to-use, simple-to-abuse" API. :+1: 

Finally, the only use case for a single child here is when you want a `<RadioSwitch />` in there - and we (or the consumer) can build a wrapper around that use case, simplifying its general use case.

selbekk:
Sounds like a face-to-face discussion sometime later this week @adidick :slightly_smiling_face: We'll discuss this later this week.

wkillerud (dismissed):
Nice work! :tada:

wkillerud (approved):

antidecaf (approved):

PR #62, opened 2018-03-07
Fix lint errors and test spam

selbekk:
Found some lint errors while working on a different branch - this pull request fixes those.


wkillerud:
Is linting not part of the CI build? In that case, shouldn't it?

wkillerud (approved):

selbekk:
Yeah it definitely should. I've created #64 to follow up.

PR #63, opened 2018-03-08
infra: Add coverage reporting to CI build (#7)

wkillerud:
This commit introduces a new test script for CI that includes coverage
reporting with Jest and Coveralls. It also uses Jests `--ci` option.
From the docs:

> When this option is provided, Jest will assume it is running in a CI
> environment. This changes the behavior when a new snapshot is
> encountered. Instead of the regular behavior of storing a new
> snapshot automatically, it will fail the test and require Jest to be
> run with `--updateSnapshot`.

Travis will use this new CI script so we get Coveralls reporting on
pull requests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
         "styleguidist:server": "styleguidist server",
         "styleguidist:build": "styleguidist build",
         "test": "jest",
+        "test:ci": "jest --ci --coverage && cat ./coverage/lcov.info | coveralls && rimraf ./coverage",
         "test:watch": "jest --watch"
     },
     "devDependencies": {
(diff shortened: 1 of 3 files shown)


wkillerud:
There we are :smile: We can configure two settings in Coveralls that can fail checks for pull requests:

1. Coverage Threshold for failure
2. Coverage Decrease Threshold for failure

We can discuss if we want to use these in issue #7 

screen shot 2018-03-08 at 09 43 17

package.json
```diff
@@ -28,6 +28,7 @@
         "styleguidist:server": "styleguidist server",
         "styleguidist:build": "styleguidist build",
         "test": "jest",
+        "test:ci": "jest --ci --coverage && cat ./coverage/lcov.info | coveralls && rimraf ./coverage",
```
selbekk: Wanna add the lint task here as well, to fix #64 ?

selbekk (dismissed):
:tada: :taco:

antidecaf (approved):

Issue #48, opened 2018-03-02
Styleguide: Flash of unstyled content when opening components page


antidecaf:
When navigating to [components](http://design.sparebank1.no/styleguidist/index.html) from a different page in the styleguide, there is a flash of unstyled content before the styling kicks in and saves the day. The problem is not present in other pages.

![screenshot from 2018-03-02 15-42-58](https://user-images.githubusercontent.com/463847/36904346-c06516a4-1e30-11e8-9c37-d8f767632a8d.png)

wkillerud:
I can take a look at this - I suppose we can move Less compilation to an `npm` script and add a `link` tag in `head` instead of bundling styles together with the JavaScript.

wkillerud:
On second thought, that would make local development more difficult. Maybe `ExtractTextPlugin`? Anyway, I'll give it a shot and open a PR for discussion.

kwltrs:
Maybe a production build, that links the two stylesheets instead of insourcing(?) them, could solve this issue. Also, having a link tag to `/styles.css` would be a slight improvement as these styles only apply to the header and navigation and would not affect local development (if one is not working on the header or navigation).

wkillerud:
Merged the improvements. They'll need to be published though, and as far as I can tell from #3 it's a manual process. @antidecaf, you fix? :bowing_man:

antidecaf:
^ @kwltrs 👼

PR #65, opened 2018-03-08
Improve experience when going to Komponenter

wkillerud:
This PR introduces a new build system for the styleguide CSS. It
lets Webpack handle it by extracting the final CSS with
`ExtractTextWebpackPlugin` if the build is a production build. We use
`NODE_ENV` to determine if we're building PROD or not, and use
`cross-env` for compatability with other platforms.

Combined with `HtmlWebpackPlugin` which comes with React Styleguidist
this setup creates a separate CSS while which is linked in `head`. This
should fix the issue of the flash of unstyled content reported in #48.

The PR also brings changes to the Handlebars templates to improve
the loading experience while we're fetching the styleguide JavaScript
bundle, which can take a while.

Finally I bring an improvement to the DX when working with Less/CSS locally
by fixing hot reloading/live changes of styles :tada:


selbekk (dismissed):
MERGE IT RIGHT AWAY

package.json
```diff
@@ -42,11 +40,13 @@
         "babel-preset-stage-3": "^6.24.1",
         "case": "^1.5.4",
         "coveralls": "^3.0.0",
+        "cross-env": "^5.1.3",
```
kwltrs, line 43: > [...] and use `cross-env` for compatability with other platforms.

Other platforms, that´s windows, or?

kwltrs (dismissed):

package.json
wkillerud: Windows, yes. Figured I could use a cross platform util since I think this is the first use of environment variables in the project.

wkillerud:
Resolved a merge conflict with the macOS fix by @antidecaf :dancing_men:

selbekk (dismissed):
great improve!

selbekk (approved):

antidecaf (approved):

PR #68, opened 2018-03-08
61 center button texts

### LINKED ISSUE #61: ffe-buttons: Text is aligned left in block level buttons
The text in block level buttons is aligned to the left, should be center. This bug is present despite both `text-align: center;` and `align-items: center;` (on the label element) being set.

screenshot from 2018-03-07 15-19-09

selbekk:
This pull request fixes a bug with ffe-buttons that doesn't center the text.

In addition, this pull request fixes an issue in the button docs that stopped us from adding a ghost modifier example.

Fixes #61

### DIFF
--- a/packages/ffe-buttons-react/src/ActionButton.js
+++ b/packages/ffe-buttons-react/src/ActionButton.js
@@ -35,6 +35,8 @@ ActionButton.propTypes = {
     disabled: bool,
     /** The rendered element, like an `<a />` or `<Link />` */
     element: oneOfType([func, string]),
+    /** Applies the ghost modifier if true. */
+    ghost: bool,
     /** Ref-setting function passed to the button element */
     innerRef: func,
     /** Shows a loader if true */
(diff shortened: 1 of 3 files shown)


packages/ffe-buttons-react/src/ActionButton.js
```diff
@@ -35,6 +35,8 @@ ActionButton.propTypes = {
     disabled: bool,
     /** The rendered element, like an `<a />` or `<Link />` */
     element: oneOfType([func, string]),
+    /** Applies the ghost modifier if true. */
+    ghost: bool,
```
wkillerud, line 39: This change isn't really related to `docs`, so I think it should either be skipped or done in its own commit.

wkillerud (changes requested):

packages/ffe-buttons-react/src/ActionButton.js
selbekk: It kind of is related though, since it changes the order this prop shows up in the docs.

wkillerud: That's a stretch, but I'll allow it :wink:

wkillerud (approved):

antidecaf (approved):

PR #70, opened 2018-03-08
chore: make build assets work on macs

antidecaf:
Fixes a bug in which `build:assets` would copy the contents of `/assets` into the root of `/dist` as opposed to `/dist/assets` on macOS


selbekk (approved):
Great job!

kwltrs (approved):

PR #74, opened 2018-03-09
Have ffe-icon resolve the path of --opt relative to current working directory

kwltrs:
Breaking changes in ffe-icons: The path passed to `--opt` is treated as relative to the current working directory, determined by `process.cwd()`.

### DIFF
--- a/packages/ffe-icons-react/package.json
+++ b/packages/ffe-icons-react/package.json
@@ -16,7 +16,7 @@
     "build:babel": "babel jsx -d lib",
     "build:jsx-components": "babel-node ./src/build-jsx-components.js",
     "clean": "rimraf tmp jsx lib",
-    "ffe:icons": "ffe-icons --opts=designsystem/packages/ffe-icons-react/svg.sprite.config.json",
+    "ffe:icons": "ffe-icons --opts=svg.sprite.config.json",
     "lint": "eslint src/.",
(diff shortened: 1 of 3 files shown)


wkillerud (dismissed):
Nice! I'd love it if the commit message contained a diff example of how to migrate. Either that or link specifically to this PR so people can see a before/after example.

wkillerud:
If you rebase from `master` the CI should be stable now (after #75)

kwltrs:
The child commit b29ee89 is a good example on how to apply this change. But I'm not sure how and where to refer to that commit.

selbekk (dismissed):
Not sure what is going on in most of this file, but I couldn't spot any major fups.

wkillerud:
Still unstable? Darn it :rage: Allright, let's just lint things in sequence. I can make a PR.
Changes LGTM :)

Regarding referecing a single commit, it looks like Github can see that b29ee89 is a commit hash and fixes a link automagically :sparkles: So just write something like

```
See b29ee89 for an example on how to migrate
```

should work in theory.

kwltrs:
When rewriting the commit with the fix, the following commit, that contains the migration, gets a new hash and the hash link is wrong. I reordered the commits so that the migration is before the fix.

wkillerud (approved):

Issue #69, opened 2018-03-08
Linting fails randomly in Travis CI


wkillerud:
Some of the linting subprocesses get killed in Travis. We may need to tweak Lerna to do fewer things in parallel during CI.

PR #75, opened 2018-03-09
fix: Limit to 2 concurrent tasks when linting

wkillerud:
We had issues with Travis CI killing random lint tasks, likely due to
the number of parallel tasks. Limit to 2 (default is 4) when on CI.

Attempted fix for #69


wkillerud:
I'll go ahead and merge this right away since randomly failing CI marks other PRs as blocked.

PR #81, opened 2018-03-09
fix(ffe-webfonts): Make is possible for frontend apps that use sb1-st…

nordstrand:
…atic-resources to get font files copied

### DIFF
--- a/packages/ffe-core/package.json
+++ b/packages/ffe-core/package.json
@@ -21,8 +21,5 @@
   },
   "publishConfig": {
     "access": "public"
-  },
-  "sb1.StaticResources": [
-    "fonts/**"
-  ]
+  }
 }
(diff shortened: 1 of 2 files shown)


wkillerud:
I see we can remove the field from `@sb1/ffe-core` now too. Would you mind removing it in a separate commit in this PR @nordstrand? :slightly_smiling_face:

kwltrs:
I feel like having a talk if we want to support the proprietary build system using this entry.

nordstrand:
Japp då har jag tagit bort detta från ffe-core @wkillerud  .

nordstrand:
@kwltrs  jag sitter hær i 2. etagen om du vill snacka?

kwltrs:
We came up with a better solution, which is not include here on github. Look up `sb1-webfonts-resources` in your local bitbucket.

PR #110, opened 2018-03-21
Reimplement ffe-system-message-react

### LINKED ISSUE #109: Modernize ffe-system-message-react
[`ffe-system-message-react`](https://github.com/SpareBank1/designsystem/blob/master/packages/ffe-system-message-react)

The current implementation has several shortcomings:

- You can't add extra classes
- It implements its own collapse instead of using `react-collapse` like the rest of FFE
- It uses the deprecated Hawaii layout system instead of ffe-grid
- It uses a div for a clickable element, which is a no-no for a11y (even with the correct roles)

To be a modern citizen!

It's old and outdated

#### Steps to reproduce
Open the source file and see for yourself: https://github.com/SpareBank1/designsystem/blob/master/packages/ffe-system-message-react/src/SystemMessage.js

selbekk:
This pull request suggests a new non-breaking re-implementation of the `ffe-system-message-react` package. This new implementation is simpler, uses the ffe grid, and uses `react-collapse` for collapsing the message once read.

Not sure if removing an optional prop is breaking - what do you guys think?

Fixes #109


wkillerud:
> non-breaking

I disagree, consumers must also have `ffe-grid` in their setup (to get the CSS), which wasn't a requirement before this commit. That seems breaking to me.

wkillerud:
> Not sure if removing an optional prop is breaking - what do you guys think?

This is also breaking. We're removing an API. Even if it's optional somebody may use it and with these commits it would stop working.

wkillerud (dismissed):
This must be marked as a breaking change.

While we're breaking things, maybe just do away with layout in this component all together? It could take up the full width of its container, then we defer layout to the consumer?

ivarni:
I'm with @wkillerud on that. We should try to avoid layout-elements in components if we can, that makes it easier to use the component and relying on grid would also make it hard to use the component for those still on hawaii layout

selbekk:
Hm... I agree fully!

Not sure how we can solve the "colored background, but aligned content" problem without using the grid though - any suggestions?

ivarni:
Is it a requirement that the content of the system message aligns itself with other content on the page or can it just have padding?

I took a quick look at http://design.sparebank1.no/styleguidist/index.html#systemmeldinger and it doesn't feel like we need a layout-system to implement that?

selbekk:
Not sure, the design requirements are vague at best. I would assume, however, that a system message typically would be at the top of the page (or right below the top header).

cc @hodoea and @adidick - do you have any usage examples we can look at when it comes to these system messages? That would be the cherry on the blossom.

wkillerud:
@selbekk, we use it in kundeforsikring.sparebank1.no to show a message if one of the core systems are down. Its background color gets the full page width and the content aligns with the main content container. 

screen shot 2018-03-22 at 10 06 31

That being said, it doesn't _have_ to look like this :)

ivarni:
That looks a lot like it could just be centered?

selbekk:
It could definitely be centered - it would solve the implementation issues. But that would be a change in design. Also, in the event of long messages - when should it wrap?

wkillerud:
> when should it wrap?

https://github.com/SpareBank1/designsystem/blob/master/packages/ffe-core/less/dimensions.less#L1-L3 @selbekk

selbekk:
Yeah that could work. That's what we base our grid on as well, IIRC. I'll come up with a simplified implementation later tonight then :) 

I'd still love some design input on this though!

selbekk:
Alright, so this was pretty simple, really. I just gave it a max-width (thanks for the tip @wkillerud), and I could do away with the entire `ffe-grid-react` dependency.

I've changed the commit message to note that the change is indeed breaking, since a few props (albeit optional ones) have been removed. Nothing's better than a soft breaking change anyways 👍

hodoea:
I dont have any examples sorry :( I´ve used native snackbars(from material design) for these kind of messages on the mobile apps. Because of maintenance issues.

What do you have in mind when you say you want design input, Kristoffer? Guidelines?

However it seems like the our Sketch library and design.sparebank1.no have two different types of designs for these messages, icons are different..  
See below:

Verktøykassen i Sketch:

skjermbilde 2018-03-22 kl 22 14 26



Design.sparebank1.no:

skjermbilde 2018-03-22 kl 22 13 02

selbekk:
Ah, we can fix that difference - I'll update the PR tonight.

Am I right in assuming that they're always going to reside at the top of the page? Or are there use cases for using system messages in context, for example inside a grid further down the page?

(edit: translated to English)

adidick-zz:
You're correct @selbekk, System/Global messages are always at the top of the page, whereas Context messages can be used in other situations throughout the page.

selbekk:
Great @adidick - should they be right below the top header, or layered on top of it?

This won't affect the code, but I think I'll write about it in the README to encourage people to use it correctly.

adidick-zz:
@selbekk Layered on top, I believe.

selbekk:
Alright! I'll change the pull request to reflect these changes as soon as possible :)

selbekk:
@hodoea I noticed that the background color for the news system message is #4b5155 - which is not in our color palette. Should I use @ffe-black or @ffe-grey-charcoal, or should we use this one-off custom color?

ivarni:
Bygg-feil:

```
less/ffe-system-message.less
 19:27  ✖  Expected "#4b5155" to be "#4B5155"   color-hex-case
 21:15  ✖  Expected "#4b5155" to be "#4B5155"   color-hex-case
```

ivarni (dismissed):
LGTM om bygget fikses

selbekk:
Oops! Will fix :) I would love to get rid of this color variation though - perhaps @hodoea or @adidick can help me out here.

hodoea:
@selbekk The example color is: 4A5155 btw(not 4B5155). I am not sure what color the other teams are using, but we have used 4A5155 in our system news messages in the login-page for the mobile bank app.

Heidi and Julia are creating new guidelines for the color palette. So they might be able to answer. I can´t tag them here btw.. :/ Do they have to be added in to the repo or something?

selbekk:
@hodoea Not completely sure why you can't tag them - perhaps you can send them the link to this conversation via IM or something?

Sorry for posting the wrong color code - I had to read it off the picture with a color meter, so I was kind of assuming it was a bit off. I don't mind adding an extra color - but it probably should be a part of the color palette, and not one-off inside this component.

adidick-zz:
#4A5155 i s a legacy colour as used by "nye nettside" project as the primary colour for navigation elements (In the main header menu). It wasn't included in the colour palette in the design system because it is a one-off use. Ideally we should either add it into the system and give it a specific use case or replace it with a similar colour from the design system. As Hodo said, it's likely something that Heidi, Julie and possibly others from the Bank team are looking into. I can add it to the agenda at our next GD forum.

selbekk:
Great! I'll leave it as it is now then (I'll fix the color code), and then we can revisit whether it needs its own variable or if it should be replaced.

adidick-zz:
@hodoea #4A5155 wasn't meant to be used anywhere else as far as I remember. It was chosen to be a more subtle black in the navigation/header. Usually we use #262626 in all other messages as it's darker and has better contrast when used over a coloured background. I'm not super familiar with what you guys have done in the mobile bank app, so maybe it was an active choice to use #4A5155 there instead of #262626?

hodoea:
@adidick Sorry I was mistaken, we actually used #676767 (grey charcoal). Probably better to keep it consistent and use #262626 if it is better for the blue login page in the app. We are going to revisit it along with several other issues :)

selbekk:
@adidick note that this color #4A5155 is a background color, not a text color. I'll merge this as it is now (with the `#4A5155` color code for the background, like the sketch example), and then we can create a patch release changing it to whatever you guys decide at a later point in time.

PR #112, opened 2018-03-22
fix(ffe-datepicker-react):fixes #97-datepicker-click-error

### LINKED ISSUE #97: ffe-datepicker navigation with incomplete date in text input field
ffe-datepicker
The datepicker month/day buttons are locked when the date in the text field is incomplete, for example when we delete a character in a year. It makes it hard to just 'click yourself' to a date in the calendar, change a month etc.
datepicker screenshot from 2018-03-15

Might be easier to just let the user click on a date instead of forcing the user to use a keyboard to fix a mistake. If the month switcher is clicked with an incomplete date, the datepicker shoud asume the relative date we are changing from is the first day in the currently displayed month so we may proceed with using the mouse for navigation.
The datepicker doesn't let the user change the month when an incomplete date is in the text input.
#### Steps to reproduce
Provide a valid date, remove last character in the year, try to change the month.

ksawerysb1:
fixed eventing issue where calendar with invalid date in input field would close before it could run onClick handler, making clicking on dates useless.

Makes the date stick to the last valid date if an invalid date is typed to the input field.


ksawerysb1:
TODO: after removing a character from the input field, making the date invalid, clicking on the datepicker date creates a validation error, then it changes the input field. It's cosmetic, but can be improved, only problem is that the onBlur eventing is tricky to implement after the focus has been lost to the datepicker.

selbekk:
fixes #97

ksawerysb1:
TODO: a mechanism for closing the datepicker after the events had been executed should be implemented, there is an issue with test:
      343 |                 openCalendarAndBlurDateInput(wrapper);
    > 344 |                 expect(wrapper.find(Calendar).exists()).toBe(false);

selbekk:
Please run your changes through `prettier` to ensure consistent formatting.

Also, would be great to see some tests that verify the behavior you're trying to fix!

selbekk (changes requested):
Please run your changes through prettier to ensure consistent formatting.

Also, would be great to see some tests that verify the behavior you're trying to fix!

ksawerysb1:
@selbekk prettiefied and added tests. 
Changed the order of elements to change the clicking area while the error message is displayed. Focus switching between input field and calendar should be now be correct when error pops up as well as when the input field is changed multiple times.
Open for suggestions about renaming or re-factoring, the element behaviour seems to be good though.

selbekk (approved):

selbekk:
Looks good to me @ksawerysb1 👍 Thanks!

Issue #107, opened 2018-03-21
ExpandButton takes full width of surrounding container when expanded 


torbein:
#### Related to:
designsystem/packages/ffe-buttons/less/base-button.less

ExpandButton when expanded is now taking the full width of the surrounding container. This behavior has been introduced in newer versions, but should should be changed back to how it was before, where the width is decided be the icon shown in the ExpandButton.

#### Expected behavior? 
Should have the same width as the icon and include transition styling. 
![expandbutton-expanded-not-full-width](https://user-images.githubusercontent.com/1019313/37709872-eef5de74-2d0b-11e8-82b0-9f6ef72a5804.png)

#### Actual behavior?
Takes the full full width of the surrounding container.
![expandbutton-expanded](https://user-images.githubusercontent.com/1019313/37709870-eedb6d64-2d0b-11e8-94d2-db3bb62fa276.png)

#### Steps to reproduce
Have screen size smaller than 480px. Press the "ExpandedButton", when it is clicked the button takes the full width of the surrounding container and shows an "X"-icon which should be having the same width as the icon.

wkillerud:
Fixed in #113

PR #113, opened 2018-03-22
fix(ffe-buttons): Set fixed width for expanded modifier

torbein:
Fixes issue #107

### DIFF
--- a/packages/ffe-buttons/less/base-button.less
+++ b/packages/ffe-buttons/less/base-button.less
@@ -108,6 +108,7 @@
 
     &--expanded {
         padding: 8px;
+        width: 45px;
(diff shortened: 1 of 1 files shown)


selbekk (changes requested):
Thanks for the contribution! 

Inline the variable instead, this unit isn't a particular candidate for re-use, so I suggest that we don't keep it in its own variable.

torbein:
Updated and set the variable inline instead.

wkillerud (approved):

selbekk (approved):

selbekk:
Great job - thanks a ton @torbein 👍

PR #119, opened 2018-03-23
chore: Add React 16.2.0 as a peerDependency

### LINKED ISSUE #44: Explicitly drop support for React 15, and other React questions
We have React 15.x listed as a valid `peerDependency` for two packages:

- `@sb1/ffe-accordion-react`
- `@sb1/ffe-tabs-react`

We use `Fragment`, a React 16.2.0 feature in two packages:

- `@sb1/ffe-details-list-react`
- `@sb1/ffe-file-uploader-react`

In neither of the latter two packages do we list React as a peerDependency (at all, nevermind 16.2.0).

There is also an inconsistency in whether we list React as a devDependency in the individual packages or if we defer this to the repo root.

There are multiple issues that should be resolved, but the most important one I think is in the issue title:

- Should we explicitly drop support for React 15?
  - If no, then should we remove the usage of `Fragment` and roll back our devDependencies?
- Should we list React as a peerDependency in all `-react` packages?
- Should we list React as a devDependency in all `-react` packages?

wkillerud:
Remove 15.x as a valid peerDependency so people get warnings to upgrade
since we may use 16.2 features such as Fragment at any time (and already
have for some components).

Fixes #44 

I'm unsure, but this may be a breaking change if npm throws errors for these unmet/wrong versions. Anybody know for sure?

BTW: I'm going on easter holiday so I won't be available for a while :)

### DIFF
--- a/packages/ffe-accordion-react/package.json
+++ b/packages/ffe-accordion-react/package.json
@@ -32,11 +32,13 @@
   "devDependencies": {
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
-    "jest": "^22.4.0"
+    "jest": "^22.4.0",
+    "react": "^16.2.0",
+    "react-dom": "^16.2.0"
   },
   "peerDependencies": {
     "@sb1/ffe-accordion": "^4.0.0",
-    "react": "^15.0.0 || ^16.0.0"
+    "react": "^16.2.0"
   },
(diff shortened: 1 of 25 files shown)


ivarni:
IIRC it should only throw errors on npm `2.x`. Starting from `3.x` unmet peer dependencies should be a warning.

kwltrs (approved):

PR #123, opened 2018-03-25
feat(ffe-tables): modifier to preserve whitespace

ivarni:
Added a `--preserve-white-space` modifier to the table cell
class `ffe-table__cell`. Using this will add `white-space: pre`
to the cell it's used on.

### DIFF
--- a/packages/ffe-tables/less/table.less
+++ b/packages/ffe-tables/less/table.less
@@ -72,6 +72,10 @@
             vertical-align: top;
         }
 
+        &--preserve-white-space {
+            white-space: pre;
+        }
+
         @media screen and (min-width: @breakpoint-md) {
             display: table-cell;
             padding: 19px 10px;


ivarni:
It doesn't seem like the less-modules are documented any more so I couldn't figure out a good place to add documentation for this modifier. Normally the React component would have it but

1) Nothing is currently documentet in the [styleguide](https://design.sparebank1.no/styleguidist/index.html#table) because there are no inline comments for anything in [the propTypes](https://github.com/SpareBank1/designsystem/blob/master/packages/ffe-tables-react/src/Table/Table.js#L114) and it looks like a giant scope-creep to have to document the entire component just to add a modifier in the less-package.

2) I don't see how the current API of `Table` would let me selectively add a modifier to a table cell in any kind of useful way so I can't really make a corresponding change in the component that can act as documentation

If nothing else, it's going to be listed in the changelog so that's something I guess.

kwltrs:
Is this a common case? If not, could this be achieved by nesting `<pre>` tags or with app-specific css classes?

ivarni:
I dunno 🤷‍♂️ 

It's a case for me right now at least. `<pre>` changes the font-family and I thought we were trying to avoid app-specific changes to the FFE css but it's actually less work for me to do that so if you prefer a local solution, sure.

selbekk:
> Nothing is currently documentet in the styleguide because there are no inline comments for anything in the propTypes 

You're looking at the [wrong prop types](https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-tables-react/src/Table.js)

As you point out, though, they aren't showing up in the style guide. That's a bug - [I've created an issue for it](https://github.com/SpareBank1/designsystem/issues/126).

As for the modifiers, it can be added to each column - wouldn't that be good enough? Like [the `alignRight` modifier found here](https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-tables-react/src/TableParts/TableCell.js#L28).

ivarni:
Uh.. it's already implemented like that but just not documented? I considered the possibility of adding visual modifiers as properties to either the `data` or `columns` arrays but decided that would break the one  single thing I like about that API, which is the separationg of data and presentation.

I'm inclined to leave the React component alone, I'm not using it, it was more a general observation that LESS components on their own don't seem documented. I just came here to add a CSS modifier. If that's not acceptable I'll just stick with the suggestion to make my own app-specific class.

selbekk:
> Uh.. it's already implemented like that but just not documented?

If you refer to the documentation, then yes. Looks like there was some kind of bug in the source code parser that Styleguidist uses that didn't recognize this component as a component. That's fixed now in [this pull request](https://github.com/SpareBank1/designsystem/pull/129). If you refer to something else, I don't think I quite understand what 😄 

The columns array is our API to add behavior and looks to each column. Examples are whether or not they are sortable, or if you should align the text a certain way. All the data is passed through the `data` prop.

_All of that aside_, I think I agree with @kwltrs here - a local solution might be the most appropriate here (at least until we see there's a lot of preformatted text in our consumer code bases.

kwltrs:
Can we close this PR?

ivarni:
That's up to you. 
I'd like to see it merged (or I wouldn't have created the PR), but it's ultimately not my decision what happens with it.

PR #129, opened 2018-03-27
docs(ffe-tables-react): Fix prop type docs

### LINKED ISSUE #126: Missing prop type documentation for <Table />
ffe-tables-react

The `<Table />` prop types are not showing up in the design guide, They should show up - but I have a feeling that since there is a folder named `Table` and a file `Table.js`, some info is lost. Perhaps a rename would fix it?

selbekk:
This commit changes the implementation of the ffe-tables-react base
class so that it works with Styleguidist and its JSDoc parser. Now there are SO MANY PROPS!

image

.... and so on - I couldn't fit them all on my screen!

This commit also translates and (veeery slightly) improves the usage
documentation.

Fixes #126


ivarni:
It seems to me that it might be an idea to have a look at all those props and see if we really need them all.

Why is there a `smallHeader` when we also have `condensed`? What's the difference between `columnLayoutMobile` and `breakpoint` (and should we really have those? why should some tables look vastly different across apps?).

antidecaf (approved):

selbekk:
Yup, I agree @ivarni - this API is pretty out of control. We should probably take a foot in the hill and consider how the API can be improved on a more general level. Passing in massaged data structures doesn't seem to fit many use cases - perhaps a more granular approach is better (where the consumer renders the rows and columns themselves). 

Either way, this pull request probably shouldn't be the arena for that discussion. I've created #132 for discussing this API further - perhaps we can continue any API discussions there? This PR is kind of unrelated to the bigger question of "should we rewrite our API?".

wkillerud (approved):

PR #138, opened 2018-04-04
feat: Add illustrations to decorate styleguide contents

antidecaf:
Adds illustrations to decorate styleguide contents. Only visible on desktop for now, and only two illustrations to begin with. (But more work is set to be done on responsive svgs and illustrations in general in the not too distant future.)

sc-front
sc-kom-i-gang


src/templates/frontpage.hbs
```diff
@@ -18,6 +18,23 @@
         <main class="sb1ds-main">
           <div class="sb1ds-intro sb1ds-intro--frontpage">
             <h1 class="ffe-h1 sb1ds-intro__heading">SpareBank 1 Designsystem</h1>
+            <div class="sb1ds-intro__illustration sb1ds-intro__illustration--frontpage" role="presentation">
+              <svg id="luftballong"
+                  xmlns="http://www.w3.org/2000/svg" viewBox="0 0 455.269 374.333">
+                  <title>Luftballong</title>
```
wkillerud: Since it's presentation only we may not need this `title`

wkillerud (dismissed):

hodoea (approved):

selbekk (approved):

adidick-zz:
@antidecaf I have some questions in regards to making the padding/margins in the top header bigger (which would possibly affect the illustrations). It's currently quite squashed up there and more space will give the illustrations more room to breathe. Can we maybe have a look at that together tomorrow?

PR #141, opened 2018-04-05
feat(ffe-form-react): Adds a PhoneNumber component to ffe-form

jenmo917:


selbekk:
Hi Jens! Thanks for wanting to contribute to our design system!

I'm not sure if these kind of compound components are something we want to have in our design system. Is there any reason why this can't live in user-land?

antidecaf:
It's a reusable implementation of a design pattern that would otherwise be implemented identically (or not) in several different apps. Imho this is certainly a helpful contribution.

selbekk:
Very well! 

In that case, I have two questions:
- should it live in the `ffe-form-react` package, or should it have its own package, like `ffe-searchable-dropdown` and `ffe-account-selector` have?
- should it not include labels and input groups? We've moved away from having these in other components, so that they're easier to reuse across applications.

Don't get me wrong, I think it's a great contribution, I'm just wondering whether this is the right place to put it.

wkillerud:
My 2 cents: I think this is a useful addition and I think it can live nicely in `ffe-form(-react)`. It's not complex enough in my opinion to warrant its own package. Seeing as it's already implemented in `form` I'd say we don't spend time extracting it :)

packages/ffe-form/CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+      <a name="9.0.5"></a>
```
wkillerud: This changelog entry and version change in `package.json` is done automatically by Lerna when we publish the version, so please don't add them manually.

The changes in `ffe-form` should be done in a separate commit so you get two commits in this PR:

1. `feat(ffe-form): Add new phone number component`
2. `feat(ffe-form-react): Add new PhoneNumber component`

packages/ffe-form-react/src/PhoneNumberField.js
```diff
@@ -0,0 +1,82 @@
...
+                    <label className="ffe-form-label" htmlFor="phone-input">{text.PHONE_NUMBER}</label>
+                    <input
+                        id="phone-input"
+                        maxLength="14"
+                        type="tel"
+                        className="ffe-input-field phone-number-field__phone-input"
+                        onChange={onPhoneChange}
+                        onBlur={onBlur}
+                        value={phone}
+                        aria-invalid={showValidationError}
```
wkillerud: We try to avoid making custom aliases for HTML properties. The component should instead apply `aria-invalid` to its children. You can see an existing example in [RadioButtonInputGroup](https://github.com/SpareBank1/designsystem/blob/36dc3a255a300b3639b727ccdef8729c24a6a036/packages/ffe-radio-button-react/src/RadioButtonInputGroup.js#L63) where the component takes `aria-invalid` and sends it to the child elements that need it.

packages/ffe-form-react/src/PhoneNumberField.js
```diff
@@ -0,0 +1,82 @@
...
+                    />
+                </div>
+            </div>
+        </div>
+    );
+}
+
+PhoneNumberField.propTypes = {
+    phone: string.isRequired,
+    countryCode: string.isRequired,
```
wkillerud: Could we have the Norwegian country code as a default property here? Not all our apps support languages other than Norwegian, especially internal ones.

packages/ffe-form-react/src/PhoneNumberField.js
```diff
@@ -0,0 +1,82 @@
...
+                            onChange={onCountryCodeChange}
+                            onBlur={handleBlur}
+                        />
+                    </div>
+                </div>
+                <div className="phone-number-field__number">
+                    <label className="ffe-form-label" htmlFor="phone-input">{text.PHONE_NUMBER}</label>
+                    <input
+                        id="phone-input"
+                        maxLength="14"
```
wkillerud: It's not written down as "law" anywhere, but the team I'm in prefer not putting such maxLengths on inputs. We instead validate the input using JavaScript.

packages/ffe-form-react/src/PhoneNumberField.js
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    function handleBlur(e) {
+        if (e.target.id !== 'country-code' || e.target.value.length === 0) {
+            onBlur(e);
+        }
+    }
+
+    return (
+        <div className={`ffe-input-group ${className ? className : 'phone-number-field'}`}>
```
wkillerud: Looks like the `phone-number-field` class is applied to two `div`s here - this one and its direct child

packages/ffe-form-react/src/PhoneNumberField.md
```diff
@@ -0,0 +1,9 @@
+Telefonnummer med landskode
+
+```js
+<PhoneNumberField 
+    phone={"123123123"} 
+    countryCode={"47"}
```
wkillerud: More a style preference, but for Strings you can skip the curly braces:

```jsx
countryCode="47"
phone="12345678"
```

packages/ffe-form-react/src/PhoneNumberField.js
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import {string, bool, func} from 'prop-types';
+
+import i18n from './i18n/i18n';
+
+export default function PhoneNumberField({
+                               phone,
```
wkillerud: Again more a style preference, but when destructuring this much I prefer doing it outside the function signature:

```js
export default function PhoneNumberField(props) {
    const {
        phone,
        countryCode,
        onChange,
        onBlur,
        locale,
        disabled,
        showValidationError,
        className,
    } = props;
```

wkillerud (changes requested):
Sorry for the comment avalanche 😅 Feel free to ignore the stuff that is down to my personal preference 😇 

The only changes I'd like to see before hitting Approve is the changelog and package.json + two commits, and maybe the showValidationError prop.

packages/ffe-form-react/src/PhoneNumberField.js
ivarni: It could probably just forward all the props it doesn't care about to the `<input>` element. That way a consumer of this component could write 
```xml
<PhoneNumberField
    aria-invalid={true}
    maxLength="14"
/>
```
And then the component wouldn't need to have opinions about maxLength either.

We could just do
```javascript
    const {
        countryCode,
        onChange,
        onBlur,
        locale,
        className,
        ...rest
    } = props;
```
And then send `rest` to the input. That way we don't have to manually pass stuff like the `disabled` prop either. `value` could also be passed this way instead of being aliased as `phone`? I feel that would be more in line with [not renaming DOM props](https://github.com/SpareBank1/designsystem/blob/develop/CONTRIBUTING.md#videresend-props).

packages/ffe-form-react/src/PhoneNumberField.js
ivarni: Just to elaborate on that it was a concern raised by UX. Supposedly it is better to let the user do the wrong thing and then telling them that it's wrong than just stopping them from doing it without any visual indication of why.

ivarni:
Out of interrest, would it make sense to have an option for hiding the country-prefix input? Or would that leave us with what is basically just a regular ffe-input? 

There are many applications in SB1 that does not ask the user for a country-code, none of ours do.

packages/ffe-form-react/src/PhoneNumberField.js
selbekk: Agree completely - but this is slightly complicated by the fact that there are two input fields. 

- Should you be able to add `aria-invalid` to the country code input? 
- Should the properties be applied to both?
- Should you be able to disable one field. but not the other?
- How do you deal with change listeners?

Finally, if we're going to have some kind of wrapping DOM element - how should we forward props to that element?

packages/ffe-form-react/src/PhoneNumberField.md
selbekk: This should be corrected by Prettier, if i remember correctly - but perhaps not in markdown files

packages/ffe-form/less/phone-number-field.less
```diff
@@ -0,0 +1,45 @@
+.phone-number-field {
```
selbekk: This component should be prefixed with `ffe-`. This helps us recognize ffe-components in client code, and it conforms with the rest of our components, of course :)

packages/ffe-form-react/src/PhoneNumberField.js
```diff
@@ -0,0 +1,82 @@
...
+PhoneNumberField.propTypes = {
+    phone: string.isRequired,
+    countryCode: string.isRequired,
+    onChange: func.isRequired,
+    onBlur: func.isRequired,
+    locale: string,
+    disabled: bool,
+    showValidationError: bool,
+    className: string
+};
```
selbekk: I would prefer it if neither of these fields were required. 

- The `phone` and `countryCode` props can be skipped, so you don't require this to be a [controlled component](https://reactjs.org/docs/forms.html#controlled-components).
- The `onChange` prop could be added further up the component tree. On our team, we use a validation technique that does just this.
- The `onBlur` prop is probably not required for all usecases.

Instead of requiring `onChange` and `onBlur` (or any function that is not required to make stuff actually work), you can pass in no-op functions in the static `defaultProps` field, like this:

```js
PhoneNumber.defaultProps = {
  onChange: f => f,
  onBlur: f => f,
};
```

This way, you can call them in event handlers, and assume that they're always callable.

packages/ffe-form-react/src/PhoneNumberField.js
```diff
@@ -0,0 +1,82 @@
...
+                            aria-invalid={showValidationError}
+                            onChange={onCountryCodeChange}
+                            onBlur={handleBlur}
+                        />
+                    </div>
+                </div>
+                <div className="phone-number-field__number">
+                    <label className="ffe-form-label" htmlFor="phone-input">{text.PHONE_NUMBER}</label>
+                    <input
+                        id="phone-input"
```
selbekk: specifying set IDs like this can be a potential problem. If you ever want to mount two of these on the same page, the labels won|'t know which field to focus once clicked!

This is an issue we've encountered before in other components. We typically create a unique ID for the component, like this:

```js
import uuid from 'uuid';

class PhoneNumber extends Component {
  id = uuid.v4();
  ...
  render() {
    return (...
      <label htmlFor={`phone-input-${id}`} />
      <input id={`phone-input-${id}`} />
      ...
    );
  }
}
```

selbekk (changes requested):
A few minor issues, but all in all pretty solid! Would be great to have complete test coverage as well before this is merged 👍

jenmo917:
Thank you guys! Please have a look at the new changes.

wkillerud (dismissed):

wkillerud:
LGTM! Nice work :tada:

packages/ffe-form-react/src/PhoneNumber.js
```diff
@@ -0,0 +1,88 @@
...
+                                        countryCodeInvalid,
+                                        numberInvalid,
+                                        className
+                                    }) {
+
+    const supportedLocales = ['nb', 'nn', 'en'];
+    const text = i18n[supportedLocales.indexOf(locale) !== -1 ? locale : 'nb'];
+
+    const numberId = uuid.v4();
+    const countryCodeId = uuid.v4();
```
ivarni: Here you'll get different ids every time you render, I'm not sure if screen readers appreciate that? It shouldn't be a problem for browser users.

ivarni:
I still think the `ffe-input-group` and the label should be left for the consumer to add. Prior experience with the dropdown component indicate that they're easier to reuse that way, but if everyone else is in favour of having it in there the rest LGTM.

packages/ffe-form-react/src/PhoneNumber.js
```diff
@@ -0,0 +1,88 @@
...
+                    <input
+                        id={numberId}
+                        type="tel"
+                        className="ffe-input-field ffe-phone-number__phone-input"
+                        onChange={onNumberChange}
+                        onBlur={onNumberBlur}
+                        value={number}
+                        aria-invalid={numberInvalid}
+                        disabled={disabled}
+                    />
```
selbekk: In order to stop adding props at some point - perhaps we can pass in props to these two input fields in their own objects? It has the downside of having a prop being an object, but it lets us spread "unknown" props. Any "unknown props outside of these two would be spread on both, making sure you could add i.e. `aria-invalid` or `disabled` props without having to repeat yourself.

In other words, an API that would look like this:

```js
<PhoneNumberInput 
  countryCodeProps={{ value: '47', className: 'extra-country-code-class' }}
  phoneNumberProps={{ value: '22225555', onChange: props.onChange }}
  aria-invalid="false" // for both input fields
/> 
```

packages/ffe-form-react/src/PhoneNumber.js
jenmo917: @selbekk do you have any feedback on this? Do I have to move them out of the render method?

packages/ffe-form-react/src/PhoneNumber.js
jenmo917: That's a refactoring we could do further on if necessary. Right now I think it's ok as is. What do you other guys think?

packages/ffe-form-react/src/PhoneNumber.js
selbekk: Well, we've pulled these out of the render method in a few other components, so it would probably be the "cleanest" thing to do.

That being said, you have two choices here:
1 - verify that this doesn't piss off the most common screen readers
2 - rewrite the component to a class component, and initialize the IDs as class fields.

I'd suggest approach number 2 ;-)

packages/ffe-form-react/src/PhoneNumber.js
selbekk: It would be a breaking change to refactor this out later though - why not make the API as flexible as possible while we're designing it? That being said, if every one else is okay with this, I'm not going to argue on about it :)

selbekk (dismissed):

wkillerud (approved):

jenmo917:
@wkillerud thanks for merging 😀 I was waiting to see if this component was playing along with some other components and use-cases I'm working on alongside this one. Hopefully it will 😅

PR #142, opened 2018-04-05
Move ffe-icons away from svg-sprite and PhantomJS

### LINKED ISSUE #24: ffe-icons: Consider creating a separate package for the SVG assets
Some consumers would only like the SVGs and use them directly in their app. Today we force these users to download PhantomJS as part of `svg-sprite`. This is a requirement to generate an SVG sprite of multiple icons at build-time. However, not all consumers want or need this feature.

Proposed solution:

- Create a new package `ffe-icons-assets` that only holds the SVGs
- Release a version of `ffe-icons` that uses `ffe-icons-assets` as its source for SVGs. If it is set as a devDependency and the SVG assets are copied and bundled in the same place as before when publishing we can avoid a breaking change.
- `ffe-icons-react` carries on, business as usual

This way we support these three use cases:

- I use React
- I don't use React and I just need one or two SVGs and don't want or need sprites
- I don't use React and I need several SVGs and should probably bundle them in a sprite

wkillerud:
Fixes #24 

This commit introduces a new tool ([svgstore](https://github.com/svgstore/svgstore)) for generating SVG sprites that isn't dependent on PhantomJS which has now pretty much been abandoned.

From what I can tell from visual inspection of the SVG file and the React result the output is 1:1. However I mark both `ffe-icons` and `ffe-icons-react` with `BREAKING CHANGE:` 💥  just to be safe 👨‍🏭


wkillerud:
screen shot 2018-04-05 at 18 49 31

packages/ffe-icons-react/src/build-jsx-components.js
```diff
@@ -1,31 +1,57 @@
...
 
+const createSvgMap = () => {
+    const map = {};
+    const iconsPath = path.join(__dirname, '..', 'node_modules', '@sb1', 'ffe-icons', 'icons');
+    fs.readdirSync(iconsPath)
+        .filter(fileName => fileName.match(/\.svg$/))
+        .forEach((fileName) => {
+            const iconPath = path.join(iconsPath, fileName);
+            const iconName = fileName.split('.svg')[0];
+            map[iconName] = String(fs.readFileSync(iconPath));
```
selbekk: if you specify the encoding, you don't have to cast it to a string:

```js
map[iconName] = fs.readFileSync(iconPath, 'utf8');
```

[(docs)](https://nodejs.org/api/fs.html#fs_fs_readfile_path_options_callback)

Edit: I see you have done this literally anywhere else 👍 is there a reason why it's different here?

packages/ffe-icons-react/src/build-jsx-components.js
```diff
@@ -1,31 +1,57 @@
...
-const camelCaseSVGProps = svgString =>
-    svgString
+const toJsx = svgString => {
+    const $ = cheerio.load(svgString, {
+        xmlMode: true
+    });
+    const svg = $('svg');
+    // React does not support namespace definitions
+    svg.attr('xmlns', null);
+    svg.attr('xmlns:svg', null);
```
selbekk, line 36: `xmlns:xlink` comes to mind - does React handle that automatically, or should we add it here as well? It's probably not a big deal, we don't use `xlink` too much with icons, but a poorly exported icon that we forgot to clean up might still have it.

selbekk (commented):
Great job!

selbekk (dismissed):

packages/ffe-icons-react/src/build-jsx-components.js
wkillerud: Nice, didn't know that. It's been a while since I wrote it so I don't remember. Probably just a coincidence.

packages/ffe-icons-react/src/build-jsx-components.js
wkillerud: That will probably break, but then again we can handle it in the future if there's some valid use case for `xmlns:xlink` in an icon.

packages/ffe-icons-react/src/build-jsx-components.js
```diff
@@ -1,31 +1,57 @@
...
 mkdirp.sync('./jsx');
 
+const createSvgMap = () => {
+    const map = {};
+    const iconsPath = path.join(__dirname, '..', 'node_modules', '@sb1', 'ffe-icons', 'icons');
+    fs.readdirSync(iconsPath)
+        .filter(fileName => fileName.match(/\.svg$/))
+        .forEach((fileName) => {
+            const iconPath = path.join(iconsPath, fileName);
+            const iconName = fileName.split('.svg')[0];
```
kwltrs, line 16: Or use [path.basename](https://nodejs.org/api/path.html#path_path_basename_path_ext):
```js
const iconName = path.basename(fileName, '.svg');
```

selbekk: Didn't know about that one!

kwltrs (approved):

PR #145, opened 2018-04-06
chore(ffe-account-selector-react): Remove old workaround

### LINKED ISSUE #18: Clarify comment about an IE11-issue in ffe-account-selector-react
This code comment mentions an issue with IE11, which is supposed to be fixed in React 16. As React 16 is out by now, the comment might need an update.
https://github.com/SpareBank1/designsystem/blob/d5bb1f7fa7b15411b707999469ee3814c3ae9a81/packages/ffe-account-selector-react/src/subcomponents/input-field/InputField.js#L14-L21

wkillerud:
We had a workaround for React 15 and IE11 that was fixed in React 16.
Now that React 16.2 is the minimum supported version we can remove the
workaround.

Fixes #18

### DIFF
--- a/packages/ffe-account-selector-react/src/subcomponents/input-field/InputField.js
+++ b/packages/ffe-account-selector-react/src/subcomponents/input-field/InputField.js
@@ -11,26 +11,10 @@ class Input extends Component {
         this.onFocus = this.onFocus.bind(this);
         this.onBlur = this.onBlur.bind(this);
 
-        /*
-     IE11 can drop characters when typing fast.
-     This is a known issue with React 15 and IE11 and will be fixed in React 16
-     See bugreport: https://github.com/facebook/react/issues/7027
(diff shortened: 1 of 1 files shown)


kwltrs (approved):

antidecaf (approved):

ivarni:
Shouldn't this be a breaking change since it requires a new major version of React in order to avoid a regression in IE11?

wkillerud:
We've had a soft requirement for React 16 for a while, and #119 made it official. So based on that this is a patch in my opinion.

ivarni:
Someone on 15.x will still get a regression the next time they bump their patch version of this without adhering to the soft requirement of using 16.x.

IMHO that's not really a good thing. At least with `<Fragment>` and its likes it's obvious that something is wrong since the app will crash while this sneaky patch will re-introduce a previously fixed localized bug 
 that is not obvious during development unless people are actively testing in IE11.

But I guess it's already merged so it's too late to do anything about it. I just figured I'd point it out.

Issue #149, opened 2018-04-09
Have we deprecated Hawaii layout or not?


ivarni:
#### Which package is this issue related to?

`ffe-core-react`

#### Describe your issue (screenshots welcome!)

I was under the impression we had deprecated the old Hawaii layout system but I see it's still in the [react core package](https://github.com/SpareBank1/designsystem/tree/develop/packages/ffe-core-react/src/layout). There's a note in the `.md` files telling consumers to "consider" using Grid instead. Haven't we deprecated the old layout system? Or is the new strategy to support both of them?

selbekk:
My discussions with @adidick has always indicated that the grid is the way of the future. As a response to that, I deprecated the Hawaii layout system. It's definitely still available, and probably will be for the foreseeable future, because migrating away from it requires a lot of manual work, including possibly redesigning parts of whatever you're migrating.

A good first goal in this deprecation has been to rid the design system itself of the Hawaii layout. It was used a few places (spinners, system messages etc), but I've rewritten those components now so that they're layout-agnostic. So step 1 done.

What do you suggest as the way forward?

ivarni:
I'd suggest a much clearer message about it being deprecated in the README for the component, a deprecation warning in the core-react component (or maybe just outright removing it from the core-react component and keep the css around?). 

Right now, [this](https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-core-react/src/layout/ContentContainer.md) reads more like a suggestion to go look at the grid than an indication that we might not support Hawaii in the future.
Having it easily available as a component without a very clear warning about its deprecation only encourages people to keep or start using it.

It's some work to migrate away from but from my experience doing it in an actual application there are very few places were you'll need a redesign.

selbekk:
Definitely agree that we can update the wording of these examples. This, when rendered with react-styleguidist, looks much less appealing since [the components themselves are deprecated](https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-core-react/src/layout/ContentContainer.js#L5-L7). As far as I can tell though, they're no longer documented in the styleguide, so this deprecation notice no longer serves much purpose.

We could remove the React components in the next major of `ffe-core-react`.

ivarni:
With the React components out of the way, should we look at possible strategies for getting rid of the LESS component? It would perhaps be great to hear from those who use it how (and if) they think a migration could be done.

ivarni:
Suggestion: 

[...]

PR #150, opened 2018-04-09
Remove deprecated components

ivarni:


kwltrs:
Can you explain why the react implementation is removed while the less is still around? Why not remove both in one breaking change?

ivarni:
Because I suspect not many people are using the react components (they're quite new) while I know several teams are using the less code (it's quite old).

People who are using this layout can keep doing it with their own React-wrappers and still postpone the move to grid while we at least not enourage it to spread further by having these wrappers in here. If we remove the less code then people are going to be stuck on the current core major for a long time. I feel that's a much bigger leap to take right now.

ivarni:
This relates to #149 btw

wkillerud (approved):

selbekk (approved):
Looks great, thanks!

PR #151, opened 2018-04-09
fix(ffe-datepicker-react): fixed issue with tabbing through inputs

ksawerysb1:
firefox would fall into a tab prison when navigating with keyboard
fixed issue with IE missing tabindex on datepicker
fixed issue with global event handler in firefox, the error message should no longer interfere with propper tab browsing
added test for rendered component references


packages/ffe-datepicker-react/src/datepicker/Datepicker.spec.js
```diff
@@ -410,6 +410,25 @@ describe('<Datepicker />', () => {
         });
     });
 
+    describe('should have element references in rendered components', () => {
+        const TestUtils = require('react-dom/test-utils');
+
+        describe('it has all references when calendar is open', () => {
```
selbekk:
```diff
-        describe('it has all references when calendar is open', () => {
+        it('has all references when calendar is open', () => {
```

packages/ffe-datepicker-react/src/datepicker/Datepicker.js
```diff
@@ -205,6 +205,21 @@ export default class Datepicker extends Component {
         }
     }
 
+    divBlurHandler(evt) {
+        if (!this._datepickerNode
+            || (evt.relatedTarget && !this._datepickerNode.contains(evt.relatedTarget))
+            || !this._datepickerNode.contains(evt.target)
+            || (this.datepickerCalendar && this.datepickerCalendar._wrapper && this.datepickerCalendar._wrapper.contains(evt.target) && evt.target.getAttribute('tabindex') !== '-1' && evt.currentTarget.getAttribute('tabindex') === '-1')
+        )
```
selbekk: Can you please explain what's happening here? It's a complex piece of code! 

Perhaps each `||` line could be pulled out into its own constant, so that it reads more like `if (isClosed || isClickOutsideDatepicker || isBlurOutsideDatepicker || {...`? It helps document the code for future developers.

packages/ffe-datepicker-react/src/datepicker/Datepicker.spec.js
```diff
@@ -410,6 +410,25 @@ describe('<Datepicker />', () => {
         });
     });
 
+    describe('should have element references in rendered components', () => {
+        const TestUtils = require('react-dom/test-utils');
```
selbekk: Can you please hoist import to the top of the test file?

Also, why would you need these test utils? We're already using Enzyme, and you should have all the query tools need available there.

selbekk (changes requested):
Great job! I'm thinking that the blur logic might be a tad bit more complicated than it need be. If it indeed need to look like that, a few more tests would be appreciated :)

wkillerud:
+1 on the comments by @selbekk 

Great job by the way! This one seemed tricky, thanks for taking it on :muscle:

selbekk:
Very true what @wkillerud said - this is not a fun bug to work on. Thanks for taking the time!

selbekk (approved):
LGTM!

PR #154, opened 2018-04-10
fix(ffe-radio-button-react): Add noop onChange to children

### LINKED ISSUE #153: ffe-radio-button-react: PropTypes complains about missing onChange on the radio buttons

`ffe-radio-button-react`

#### Describe your issue

Sometimes this happens when using `RadioButtonInputGroup`:

```
Warning: Failed prop type: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`.
    in input (created by BaseRadioButton)
    in BaseRadioButton (created by RadioButton)
    in RadioButton (created by RadioButtonInputGroup)
    in fieldset (created by RadioButtonInputGroup)
    in RadioButtonInputGroup 
```

Only the field is not read-only since I have an `onChange` and `selectedValue` on `RadioButtonInputGroup` - the state is handled just fine.

No propType warning

I get errors in my console

wkillerud:
RadioButtonInputGroup now sends a noop `onChange` function to its
children to silence React propTypes intermittently complaining about it
being a controlled component without an onChange - which is a lie since
we listen for the changes in the input group.

Fixes #153


packages/ffe-radio-button-react/src/RadioButtonInputGroup.js
```diff
@@ -31,6 +31,7 @@ const RadioButtonInputGroup = props => {
         'aria-invalid': ariaInvalid || String(!!fieldMessage),
         inline,
         name,
+        onChange: f =>
```
selbekk: so close!

selbekk (changes requested):

packages/ffe-radio-button-react/src/RadioButtonInputGroup.js
wkillerud: Oh for...

selbekk (approved):

selbekk (approved):

PR #156, opened 2018-04-10
feat(ffe-accordion-react): Make AccordionItems controllable

### LINKED ISSUE #155: Feature request: Make ffe-accordion-react controllable
ffe-accordion-react

I would like to close an accordion item from the "outside". Right now, I can only set the initial state with the `isOpen` property - this should probably control the internal state if it's set.

That `isOpen` controls the state if it's set.

That `isOpen` only controls the initial state.

#### Steps to reproduce
```js
class Example extends Component {
  state = { isOpen: true };
  onToggleItem = () => this.setState({ isOpen: !this.state.isOpen });
  render() { 
    return (
      <AccordionItem isOpen={this.state.isOpen} onClick={this.onToggleItem}>
        ...
        <button onClick={this.onToggleItem)>Close this item</button>
      </AccordionItem>
    );
  }
}
``` 
This won't close the item when you click the `Close this item` button

selbekk:
BREAKING CHANGE: The AccordionItem will now be a controlled component
if the `isOpen` prop is set. This means that it will only open or
close if the `isOpen` prop changes. Previously, the isOpen prop only
controlled the initial isOpen state of an AccordionItem component.

To migrate to this version, check if you're using the `isOpen` prop
anywhere. If you are, please ensure that the `isOpen` prop is controlled
from the consumer's side. You can react to clicks on the accordion item
header with passing a event listener to the `onToggle` prop.

BREAKING CHANGE: The `onOpen` and `onClose` props are removed from the
`AccordionItem` component, and replaced with a similar
`onToggle(nextIsOpen)` prop. `nextIsOpen` will be `true` if the
accordion item _was_ closed, but _will now be_ open, and will be `false`
if the accordion item _was_ open, but _will now be_ closed.

Fixes #155.


wkillerud:
Suggestion: support initial open state for an uncontrolled component with `defaultIsOpen`

Edit: Like Ivar mentions further down we have apps with double-digit accordions that work just fine being uncontrolled. We don't want to _have_ to manage their state.

packages/ffe-accordion-react/src/AccordionItem.js
```diff
@@ -11,77 +11,53 @@ const createClasses = (baseClass, isOpen, type) =>
...
-        this.onClickEnterAndSpace = this.onClickEnterAndSpace.bind(this);
+    getIsOpen() {
+        if (this.props.isOpen !== undefined) {
+            return this.props.isOpen;
+        }
+        return this.state.isOpen;
     }
 
-    onClick(e) {
+    onClick = e => {
```
ivarni: We're using this eye-breaking syntax now? 😢

ivarni:
Why do we need a breaking change and make this always be controlled to fix that issue? Can't `isOpen` still set initial state but also change current state *if* the consumer changes it?

packages/ffe-accordion-react/src/AccordionItem.js
```diff
@@ -11,77 +11,53 @@ const createClasses = (baseClass, isOpen, type) =>
     });
 
 class AccordionItem extends Component {
-    constructor(props) {
-        super();
-        this.state = {
-            isOpen: props.isOpen || false,
-        };
+    state = {
+        isOpen: this.props.isOpen || false,
```
wkillerud: `this.state.isOpen` is never read if `this.props.isOpen` is set, so this can just default to `false` in all cases.

packages/ffe-accordion-react/src/AccordionItem.js
```diff
@@ -11,77 +11,53 @@ const createClasses = (baseClass, isOpen, type) =>
...
+    };
 
-        this.onClick = this.onClick.bind(this);
-        this.onClickEnterAndSpace = this.onClickEnterAndSpace.bind(this);
+    getIsOpen() {
```
wkillerud: `getIsOpen = () => {`?

packages/ffe-accordion-react/src/AccordionItem.js
```diff
@@ -155,8 +141,7 @@ AccordionItem.propTypes = {
 AccordionItem.defaultProps = {
     hasNestedCollapse: false,
     ignoredNodeNames: [],
-    onClose: f => f,
-    onOpen: f => f,
+    onToggle: f => f,
```
wkillerud: `Close` and `Open` give better hints about behavior than `Toggle` in my opinion. Can this function name mention what is being toggled?

wkillerud: Maybe it's a given. Hmm. I'll let you decide :wink:

packages/ffe-accordion-react/src/AccordionItem.spec.js
```diff
@@ -54,27 +62,25 @@ describe('<AccordionItem />', () => {
             }); // esc
             expect(wrapper.state('isOpen')).toBe(false);
         });
-        it('toggling triggers onOpen and onClose props', () => {
-            const onOpenSpy = sinon.spy();
```
wkillerud: Travis correctly points out `sinon` is no longer used after this, so the import should be removed.

wkillerud (changes requested):

wkillerud:
> Why do we need a breaking change and make this always be controlled to fix that issue? Can't isOpen still set initial state but also change current state if the consumer changes it?

Hmm, valid question.

My immediate thought was that it introduces a risk of state getting out of sync. If a user clicks a button inside the component and the consumer isn't listening for `onClose`-events and updating its own state there's a problem. Then again, if you're trying to control its state from outside then why would you ever _not_ listen to those events.

My only argument would be to keep the API similar to controlled inputs with `defaultStuff` and `stuff` for uncontrolled vs controlled - that API might be less surprising for consumers.

ivarni:
I think it would be great to be able to use this component without setting up my own clickhandlers and state just to manage its open-state. Especially if I'm dealing with a lot of them at the same time. We have FAQs with 15-20 accordions in them...

packages/ffe-accordion-react/src/AccordionItem.js
selbekk: You mean like `onToggleOpen`? Sure.

packages/ffe-accordion-react/src/AccordionItem.spec.js
selbekk: Gah my bad! Thanks!

packages/ffe-accordion-react/src/AccordionItem.js
selbekk: We've used it several other places. It lets us skip having a constructor in most cases, which is nice 👍

ivarni: But we miss out on the explicit `.bind` which is a nicer thing...

I guess since it's in the language now we'll have to try to get used to seeing it.

selbekk:
> @ivarni I think it would be great to be able to use this component without setting up my own clickhandlers and state just to manage its open-state. 

This is only a breaking change for people that has set the `isOpen` prop. If you don't set this prop, this component will still keep track of its own `isOpen` state.

> @ivarni Why do we need a breaking change and make this always be controlled to fix that issue? Can't isOpen still set initial state but also change current state if the consumer changes it?

Good question! If you've got an initially opened item like this:

```js
<AccordionItem isOpen={true} />
```

, and then close it by clicking the top header - how are you supposed to open it from the parent component? You'd then have to set the `isOpen` prop to false, and then back to `true` in order for a `componentDidUpdate` or `getDerivedStateFromProps` function to pick up this "command".

If `initially open, but still uncontrolled` is a very common usecase, I can create a `openByDefault` like @wkillerud suggested. Would that be a better approach? It would still be breaking, but at least it would be a much easier migration path for those that have initially open accordion items.

wkillerud:
@selbekk ~~`defaultIsOpen` instead of `openByDefault` please, to follow the established convention in the ecosystem~~ Nevermind, I see the comment was just a typo

wkillerud (dismissed):

ivarni:
I wouldn't even consider using this component without having the option to use it uncontrolled exactly because I wouldn't want to maintain 15 flags in my parent-state for the one place where I could use it so I think it's a common enough usecase. If it's how the component already is working today that also points to it being a usecase people have.

Could I suggest we look to how React does controlled/uncontrolled `<input>`s and call them `open` and `defaultOpen` as a mirror to `value` and `defaultValue` and behaving the same way when it comes to if the component is controlled or not?

ivarni:
Another thing, would it save some potential overhead to have these discussions *before* we get to the PR-stage, i.e. in the issue?

selbekk:
> @ivarni [...] I wouldn't want to maintain 15 flags in my parent-state [...]

That was not my intention - I wasn't aware that we used accordions in that way. With the `default[Is]Open` prop, you won't have to do that.

> @ivarni Could I suggest we look to how React does controlled/uncontrolled <input>s and call them open and defaultOpen as a mirror to value and defaultValue and behaving the same way when it comes to if the component is controlled or not?

So a rename from `isOpen` / `defaultIsOpen` to `open` / `defaultOpen`? I don't mind that - we don't really have any standard across the code base when it comes to naming boolean props. Buttons have `isLoading`, expanded stuff has `isExpanded` etc, while grids have `center`, `middle` etc.

Otherwise, you're okay with the approach I've implemented now?

> @ivarni [...] would it save some potential overhead to have these discussions before we get to the PR-stage, i.e. in the issue?

Perhaps on larger changes, but this isn't that big of a change, and I find it easier to discuss implementation details when we're looking at code. That being said, I can ask your opinion in the issue next time.

selbekk:
Renamed the props `isOpen` and `defaultIsOpen` to `open` and `defaultOpen`, respectively.

ivarni:
I wasn't specificallty requesting you ask my opinion, I'm just asking if it would be better to discuss how to solve an issue before spending time doing it 🤷‍♂️

selbekk:
Alrighty! Anyhow, does this look good to go now?

packages/ffe-accordion-react/src/Accordion.md
```diff
@@ -29,7 +29,7 @@ til å begynne med (man kan fortsatt lukke det):
 const { WhiteAccordion, AccordionItem } = require('.');
 
 <WhiteAccordion>
-    <AccordionItem isOpen={true} title="Åpen!">
+    <AccordionItem defaultIsOpen={true} title="Åpen!">
```
ivarni: `defaultOpen` so it's in line with the implementation?

packages/ffe-accordion-react/src/AccordionItem.js
```diff
@@ -4,94 +4,70 @@ import { Collapse } from 'react-collapse';
...
-            isOpen: props.isOpen || false,
-        };
+    state = {
+        open: this.props.defaultOpen,
+    };
 
-        this.onClick = this.onClick.bind(this);
-        this.onClickEnterAndSpace = this.onClickEnterAndSpace.bind(this);
+    getOpen = () => {
+        if (this.props.open !== undefined) {
```
ivarni, line 19: I'm unsure if this `if`-block is needed.

packages/ffe-accordion-react/src/Accordion.md
selbekk: argh, thanks!

packages/ffe-accordion-react/src/AccordionItem.js
ivarni: Ah wait never mind it's `this.props` vs `this.state`.

ivarni (dismissed):

PR #157, opened 2018-04-10
fix(ffe-buttons): Fix issue with back button top padding

OptimusCrime:
The markup for inline `a`-tag back button have some weird top padding that results in unaligned text and icon. This only happens if the tag is an `a`-tag (and not `button`), as seen below.

```html
<a class="ffe-inline-button ffe-inline-button--back">
    <span class="ffe-inline-button__label">Tilbake til innstillinger</span>
</a>
```

Simply removing the top padding solves the problem. I am not sure why the padding is there. Perhaps some legacy code or something?

**Before:**

back_before

**After:**

back_after

### DIFF
--- a/packages/ffe-buttons/less/inline-base-button.less
+++ b/packages/ffe-buttons/less/inline-base-button.less
@@ -25,10 +25,6 @@
         line-height: 26px;
         padding-left: 1em;
 
-        a& {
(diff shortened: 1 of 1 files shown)


wkillerud (approved):
Probably some workaround for older code that went unchanged when the base component was changed. Nice find, thanks!

PR #161, opened 2018-04-11
Increase test coverage for various packages

wkillerud:
Increase test coverage. Should make Coveralls happy :purple_heart: 

The new test files are there to cover the actual export and props sent to the base component (such as the various FieldMessages).

I simply ignore the `NODE_ENV` blocks and their `else` cases.

### DIFF
--- a/packages/ffe-buttons-react/src/ExpandButton.spec.js
+++ b/packages/ffe-buttons-react/src/ExpandButton.spec.js
@@ -1,7 +1,10 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 
+import BamseIkon from '@sb1/ffe-icons-react/lib/bamse-ikon';
+import BestikkIkon from '@sb1/ffe-icons-react/lib/bestikk-ikon';
 import KryssIkon from '@sb1/ffe-icons-react/lib/kryss-ikon';
+
 import ExpandButton from './ExpandButton';
 
 const defaultProps = {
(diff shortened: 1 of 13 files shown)


henrikhermansen (dismissed):
LGTM! Great work! :tada:

packages/ffe-buttons-react/src/ExpandButton.spec.js
```diff
@@ -22,11 +25,28 @@ describe('<ExpandButton />', () => {
         const wrapper = getWrapper({ 'aria-label': 'some label' });
         expect(wrapper.props()).toHaveProperty('aria-label', 'some label');
     });
+    it('remders leftIcon and rightIcon', () => {
```
ivarni: *renders

packages/ffe-grid-react/src/utils.js
```diff
@@ -1,3 +1,4 @@
+/* istanbul ignore if: I don't see any good way of getting around NODE_ENV */
```
ivarni, line 1: I suppose we *could* have two test-suites where one is run (ran?) with NODE_ENV set to "production" but it seems a lot of work for little gain so I'm all for just skipping these.

ivarni (dismissed):

packages/ffe-grid-react/src/utils.js
wkillerud: Yeah, that was my conclusion as well - not worth the effort for the low amount of testable code.

selbekk:
Does Coveralls only build the develop branch? I don't see an increase in test coverage?!

selbekk (approved):

ivarni:
What's the develop branch?

selbekk:
@ivarni it's our "next" branch.

- develop: master + merged pull requests.
- master: whatever is released

develop is merged into master very 30 minutes or so, iirc. @kwltrs can probably elaborate more

PR #162, opened 2018-04-11
feat(ffe-checkbox-react): Add optional prop stopPropagation

### LINKED ISSUE #158: Unable to stop onClick event from bubbling from Checkbox
ffe-checkbox-react

Checking the checkbox with Spacebar triggers an "unstoppable" onClick-event.
This does not happen when clicking the checkbox with the mouse pointer.
This is an issue because I have the checkbox inside another clickable area, whose onClick listener is triggered by the Checkbox.

While current behavior could be argued as expected, I want to be able to stop the onClick-event from bubbling from the component.

I've tried adding onKeyUp, onKeyDown, onKeyPress and onClick to the Checkbox, all with the function `e => e.stopPropagation()`.
While the three first event handlers has no impact on the situation, the onClick handler flips the situation: checking with Spacebar is now fine, but clicking with the mouse pointer makes the onClick-event bubble.

#### Steps to reproduce
```javascript
<div onClick={ alert('you clicked me') }>
  <Checkbox />
</div>
```

#### Is there something you can do in your own code to mend the problem?
I tried wrapping the Checkbox in a span with an onClick handler.
While this sorta works, my span's size is dictated by the label inside Checkbox.
image
This means that the area outlined by the span is unclickable, which conflicts with the clickable surface I have underneath the Checkbox.

henrikhermansen:
This commit adds a new optional prop to ffe-checkbox-react. When this is
set to `true` the internal `<input>` and `<label>` of the Checkbox will
stop propagation of their onClick events.

Resolves #158

### DIFF
--- a/packages/ffe-checkbox-react/src/Checkbox.js
+++ b/packages/ffe-checkbox-react/src/Checkbox.js
@@ -1,12 +1,19 @@
 import React from 'react';
-import { bool, node, string } from 'prop-types';
+import { bool, node, string, func } from 'prop-types';
 import { v4 as hash } from 'uuid';
 import classNames from 'classnames';
 
 export default function CheckBox(props) {
-    const { children, inline, invalid, label, noMargins, ...rest } = props;
+    const { children, inline, invalid, label, noMargins, onClick, stopPropagation, ...rest } = props;
 
     const id = props.id || `checkbox-${hash()}`;
+
+    const onClickInput = event => {
+        if (stopPropagation) {
+            event.stopPropagation();
+        }
+        onClick(event);
+    };
 
     return (
         <span>
@@ -15,6 +22,7 @@ export default function CheckBox(props) {
                 id={id}
                 type="checkbox"
                 aria-invalid={String(invalid)}
+                onClick={ onClickInput }
                 {...rest}
             />
             <label
@@ -24,6 +32,8 @@ export default function CheckBox(props) {
                     'ffe-checkbox--no-margin': noMargins,
                 })}
                 htmlFor={id}
+                role="presentation"
+                onClick={ event => { if (stopPropagation) { event.stopPropagation() } } }
             >
                 {label || children}
             </label>
@@ -49,9 +59,14 @@ CheckBox.propTypes = {
     invalid: bool,
     /** The label for the checkbox */
     children: node,
+    onClick: func,
+    /** Stops onClick from bubbling from the checkbox/label */
+    stopPropagation: bool,
 };
 
 CheckBox.defaultProps = {
     inline: true,
     invalid: false,
+    onClick: () => {},
+    stopPropagation: false,
 };
--- a/packages/ffe-checkbox-react/src/Checkbox.md
+++ b/packages/ffe-checkbox-react/src/Checkbox.md
@@ -38,7 +38,7 @@ Du kan merke at et felt er ugyldig ved å sette `aria-invalid="true"`:
 </CheckBox>
 ```
 
-Du kan kan sende inn en callback-funksjon som blir kalt hver gang verdien i checkboxen endrer
+Du kan sende inn en callback-funksjon som blir kalt hver gang verdien i checkboxen endrer
 seg med `onChange`:
 
 ```js
@@ -48,6 +48,20 @@ seg med `onChange`:
 >
     Trykk for å lære litt om meg
 </CheckBox>
+```
+
+Du kan forhindre at click-event bobler opp fra checkboxen.
+Dette kan være nyttig dersom området rundt checkboxen er klikkbart.
+
+```js
+<div style={ { background: '#ccc', width: '150px' } } onClick={ () => alert('Jeg skal ikke komme frem når du bruker checkboxen') }>
+    <CheckBox
+        name="clicked"
+        stopPropagation
+    >
+        Trykk utenfor label for en alert
+    </CheckBox>
+</div>
 ```
 
 Komponenten videresender alle udokumenterte props til `<input />`-elementet.
--- a/packages/ffe-checkbox-react/src/Checkbox.spec.js
+++ b/packages/ffe-checkbox-react/src/Checkbox.spec.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { shallow, mount } from 'enzyme';
 
 import Checkbox from './Checkbox';
 
@@ -105,5 +105,23 @@ describe('<Checkbox />', () => {
             'false',
         );
         expect(wrapper.find('input').prop('tabIndex')).toBe(-1);
+    });
+
+    it('should stop event propagation when stopPropagation is true', () => {
+        const onClickOfWrapper = jest.fn();
+        const onClickOfCheckbox = jest.fn();
+        const wrapper = mount(
+            <div onClick={onClickOfWrapper} role="presentation">
+                <Checkbox
+                    onClick={onClickOfCheckbox}
+                    stopPropagation={true}
+                />
+            </div>
+        );
+
+        wrapper.find('div').find('input').simulate('click');
+
+        expect(onClickOfWrapper).toHaveBeenCalledTimes(0);
+        expect(onClickOfCheckbox).toHaveBeenCalled();
     });
 });


henrikhermansen:
As for why I added `role="presentation"` to the `<label>`: I got an error from eslint once I added the `onClick` handler. From what I could understand from [the documentation of that rule](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-noninteractive-element-interactions.md) this seems to be an acceptable solution. They have an example of [something similar](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-noninteractive-element-interactions.md#case-this-element-is-catching-bubbled-events-from-elements-that-it-contains).

ivarni:
I'm not 100% sure if having `role="presentation"` on a `<label>` is a good idea, won't that stop screen readers from reading out the label? Maybe disabling the eslint rule could be a better option here?

packages/ffe-checkbox-react/src/Checkbox.js
```diff
@@ -1,20 +1,28 @@
...
 import { v4 as hash } from 'uuid';
 import classNames from 'classnames';
 
 export default function CheckBox(props) {
-    const { children, inline, invalid, label, noMargins, ...rest } = props;
+    const { children, inline, invalid, label, noMargins, onClick, stopPropagation, ...rest } = props;
 
     const id = props.id || `checkbox-${hash()}`;
 
+    const onClickInput = event => {
```
wkillerud, line 11: I disagree that this should be a feature we implement inside the component. `onClick` is passed in by the consumer and they can stop propagation if they need to in their own click handler.

```js
<CheckBox 
    name="interests" 
    onClick={(event) => {
        event.stopPropagation();
        // do stuff
    }}
    value="food"
>
    Matlaging
</CheckBox>
```

If they only need to stop it if the label is creating the event then they can use `event.target.tagName` to check:

```js
<CheckBox 
    name="interests" 
    onClick={(event) => {
        if (event.target.tagName.toLowerCase() === 'label') {
            event.stopPropagation();
        }
        // do stuff
    }}
    value="food"
>
    Matlaging
</CheckBox>
```

packages/ffe-checkbox-react/src/Checkbox.js
```diff
@@ -24,6 +32,8 @@ export default function CheckBox(props) {
                     'ffe-checkbox--no-margin': noMargins,
                 })}
                 htmlFor={id}
+                role="presentation"
+                onClick={ event => { if (stopPropagation) { event.stopPropagation() } } }
```
wkillerud, line 36: How is `this.props.onClick` run in this case?

wkillerud (changes requested):
Maybe I'm overlooking something with the `stopPropagation` prop, but I don't think it should be part of the API for this component.

packages/ffe-checkbox-react/src/Checkbox.js
henrikhermansen: If this had worked, it would've been great. I mentioned in the issue that I've tried this, but I can explain a bit closer here.
When you click on the checkbox or the label, you actually get a click-event bubbling from tag name `input` (I'm catching this with `onClick` in a surrouning `span`).
When I try to stop it (regardless of target tag name):
```javascript
<Checkbox onClick={ event => event.stopPropagation() } />
```
I still get a bubbling click-event. The only difference is that now it has target tag name `label`.
I'm not sure exactly how the events spawn and bubble when you have a label and an input, but so far the only way I've found of stopping it is to stop propagation from both the `input` and the `label`.

If you have any other suggestions, I'd be happy to explore them!

ivarni: is `stopImmediatePropagation()` doing better work?

wkillerud: > When I try to stop it (regardless of target tag name) I still get a bubbling click-event. The only difference is that now it has target tag name label.

The tag name is changed when you stop propagation? Wat.
Maybe [stopImmediatePropagation](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopImmediatePropagation) like Ivar suggests would work? 

> If several listeners are attached to the same element for the same event type, they are called in order in which they have been added. If during one such call, event.stopImmediatePropagation() is called, no remaining listeners will be called.

The behavior you describe could be down to the two elements being involved. On the other hand it may be two sources, not two listeners so YMMV.

If none of this works I think you should roll your own version of `Checkbox` to be honest. It's not a very complex component and I don't like this edge case API being there for all consumers, the vast majority of which won't need this.

henrikhermansen: `Uncaught TypeError: event.stopImmediatePropagation is not a function`
Does that happen if there's only one listener?

I see I misinterpreted the events a bit before. I dug a bit more and see more clearly what happens now.

When you click on the label, you actually get two click-events bubbling. One from the `input` and one from the `label`. This is why it seemed like the expandable row I mentioned in the issue didn't expand to begin with. The truth is it both expanded and collapsed.
Focusing the label and pressing space (to toggle the box) only triggers a click event from the `input`, and thus the row is expanded/collapsed.

When you add
```javascript
<Checkbox onClick={ event => event.stopPropagation() } />
```
you do in fact stop the click event from the `input`. So now focusing and using spacebar is fine - there's no event bubbling. However, the event from the `label` is still bubbling, and thus the row is expanded/collapsed.

It still seems to me that you need to explicitly stop propagation of both the `label` and the `input` click events.

packages/ffe-checkbox-react/src/Checkbox.js
henrikhermansen: It runs just fine. Clicking the `label` also triggers a click event on the `input`. Not because of event bubbling - these are two separate events triggered because of the link between them through `id` and `htmlFor`.

I tried writing a test for this, but clicking the `label` with `enzyme` doesn't trigger the event on the `input`. This is simply a limitiation of `enzyme`.

packages/ffe-checkbox-react/src/Checkbox.js
ivarni: Right, it's probably a synthetic React-event and not a native event so it most likely doesn't have stopImmediate as part of its API.

henrikhermansen: > If none of this works I think you should roll your own version of Checkbox to be honest. It's not a very complex component and I don't like this edge case API being there for all consumers, the vast majority of which won't need this.

This is indeed an option, and I could easily copy the work I did here to our own code.
However, I do feel it's nice to be able to stick to the ffe components.

This is sort of the purpose of these components - that they should be useful for everyone, so that we can reduce the number of custom components going around.

I do agree we will always need to have some custom components, but I believe it's a weighing of complexity of the solution and the demand of the solution. The higher demand, the higher complexity we could implement. The complexity of this solution, however, is rather low. And we don't really know how the demand is.

This all being said, I'm not overly opposed to making our own component. But I'd sure like a couple more opinions on this matter.

wkillerud: Re TypeError: We have entered [React synthetic events](https://reactjs.org/docs/events.html) vs [native events land](https://developer.mozilla.org/en-US/docs/Web/API/Event/Event). Here there be monsters :skull_and_crossbones: Apparently you can reach the native event through `event.nativeEvent`, so `event.nativeEvent.stopImmediatePropagation()` should not produce the error.

henrikhermansen: > `event.nativeEvent.stopImmediatePropagation()` should not produce the error.

Thanks for the clearification. This does indeed not produce the error, but it doesn't help stopping the click event bubbling from the `label`.

selbekk:
[Here's a reproducable example with a checkbox inside a label inside a button inside of a clickable span inside of a clickable div](https://codesandbox.io/s/34nn0r3vz6) (without the FFE implementation, but this is a challenge with the way the DOM handles events, not this component specifically). Open the console tab to see what's being triggered.

Basically, if you wrap the checkbox with a `<span onClick={e => e.stopPropagation()} />`, any outside click listeners won't be affected by it. So let's say you have a checkbox inside of an accordion, you can do something like this:

```js
const Iterim = props => (
  <span onClick={e => e.stopPropagation()}>
    <Checkbox name={props.name} />
    {props.children}
  </span>
);
<Accordion>
  <AccordionItem 
    heading={<Iterim name="checked">Accordion item heading</Iterim>}
  >
    {....}
```

This way, no change is required in the component!

henrikhermansen:
@selbekk as I explained in issue #158 this is not an acceptable solution, because the span you wrap around it will have a forced size from the Checkbox, and thus create an unclickable area outside the checkbox.

I had a closer look at the cause for this now. I'm not enough of a CSS expert to fully explain it, but it makes a difference whether the Checkbox is inline or not.
Default inline=true sets `display: inline-block` on the `label` inside the `Checkbox`,
while inline=false sets `display: block` on the `label`.
These settings makes a difference to the surrounding `span`. When label is a block, the span disappears behind the Checkbox, but when the label is an inline-block the span receives its forced size from the margins of the label.

I fully agree it would be a much better solution to somehow combat this issue with the styling, but so far I haven't found a good way of doing that.

If we can't solve the styling here, my opinion is still to apply this change to the API.

PR #165, opened 2018-04-13
fix(ffe-header): Extend all necessary modifiers in order to render logout button correctly

antidecaf:
Fixes a bug in which the logout button didn't extend all necessary classes/modifiers, and therefore was rendered with only parts of the styling needed (Font-family, hover/focus/active states, etc)

screenshot from 2018-04-13 10-59-54

### DIFF
--- a/packages/ffe-header/less/ffe-header.less
+++ b/packages/ffe-header/less/ffe-header.less
@@ -135,10 +135,14 @@
         &:extend(.ffe-button--condensed);
 
         margin: 8px 10px;
+        border-radius: 20px;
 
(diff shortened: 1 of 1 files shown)


wkillerud (approved):
It seems to me this button doesn't have much unique styling. Can we maybe consider just adding a `--rounded` modifier to the secondary button with guidelines on where to use it? 

An alternative is to just copy the actual styles over from `buttons`. That way changes in `buttons` don't have unintended side effects in this special logout button.

antidecaf:
@wkillerud Definitely sounds like a good option.

PR #182, opened 2018-04-18
 fix(ffe-buttons): Properly align tertiary button anchors 

### LINKED ISSUE #178: Tertiary button text and border not aligned properly in the button group

`ffe-buttons`

#### Describe your issue

From [design.sparebank1.no](https://design.sparebank1.no/styleguidist/index.html#buttongroup):

screen shot 2018-04-17 at 09 57 13

Tested in both Firefox and Chrome.

The button text and border should be closer together like outside the button group, and vertically centered.

screen shot 2018-04-17 at 09 56 37

wkillerud:
We had a bug where the border and label of the tertiary button would be
rendered at the top and bottom of the button group if the button was an
anchor element (or !button, really). This commit adds the proper flex
attibute to make sure the content inside the container is vertically
centered. It also adjusts the CSS for the ::after element now that the
container takes care of the details.

Fixes #178


henrikhermansen (approved):
I experience the same problem now that I'm bumping all our ffe-packages.
These style changes fixes the issue. Good job!

Issue #173, opened 2018-04-16
`ffe-button-group` no longer aligns with content in `ffe-content-container`


wkillerud:
#### Which package is this issue related to?

`ffe-buttons`

#### Describe your issue

`ffe-button-group` no longer centers in `ffe-content-container` after the transition to `display: flex;` (introduced in version `7.0.2`). Before the change `ffe-button-group` had `display: inline-block;` and would be centered in the default `ffe-content-container`. Now it is left-aligned, and there is no way of centering it. The `--inline` modifier has no equivalent on large screens in the current version, so using it does nothing.

#### What is the expected behavior? 

`ffe-button-group` should be aligned by `ffe-content-container`, both its default center and its `--text-left` variant.

#### What is the actual behavior?

`ffe-button-group` is always left aligned

#### Steps to reproduce

```html
<div class="ffe-section-wrapper">
    <div class="ffe-content-container">
        <div class="ffe-button-group">
            <button class="ffe-button ffe-button--action"><span class="ffe-button__label">Klikk på meg!</span><div aria-hidden="true" aria-label="Vennligst vent" class="ffe-button__spinner"></div></button>
        </div>
    </div>
</div>
```

wkillerud:
So, is this a `wontfix` given the discussions in #149 and #150?

A workaround is to use `ffe-grid` for the layout around the button group, or to not use `ffe-button-group` at all.

wkillerud:
Might just be a matter of setting `display: inline-flex;` instead of `display: flex;` on `ffe-button-group` (I initially misread where it was set). It looks OK in an app that uses the old Hawaii layout, and looks OK in the styleguide. I'll open a PR.

PR #187, opened 2018-04-19
fix(ffe-buttons): button group direction on small screen

### LINKED ISSUE #114: ffe-button-group do not wrap buttons onto multiple lines
#### Package this issue related to:
designsystem/packages/ffe-buttons/less/button-group.less
#### Issue:
As shown in the screenshot below the buttons in ffe-button-group div do not wrap when the width of the parent is to small. 
screenshot from 2018-03-22 12-10-46

#### Expected behavior:
Buttons that have ffe-buttons-group as parent should wrap when the width is to small.
#### Actual behavior:
Buttons that have ffe-buttons-group as parent do not wrap when width is to small.
#### Steps to reproduce
Go to http://design.sparebank1.no/styleguidist/index.html#buttongroup and resize the screen to 480px in width.

henrikhermansen:
This commit changes the styling of `ffe-button-group` on small screen
(below @breakpoint-sm): our full-width buttons now wraps vertically,
as intended.

Fixes #114

### DIFF
--- a/packages/ffe-buttons/less/button-group.less
+++ b/packages/ffe-buttons/less/button-group.less
@@ -14,12 +14,12 @@
 //
 // Styleguide ffe-buttons.button-group
 .ffe-button-group {
-    display: inline-flex;
+    display: flex;
+    flex-direction: column;
     padding: 40px 0;
 
     .ffe-button,
     .ffe-inline-button {
-        flex-direction: column;
         justify-content: center;
         margin: 0 auto 10px;
     }
@@ -42,10 +42,11 @@
     }
 
     @media screen and (min-width: @breakpoint-sm) {
+        display: inline-flex;
+        flex-direction: row;
         .ffe-button,
         .ffe-inline-button {
             display: inline-flex;
-            flex-direction: row;
             margin: 0 0 10px 10px;
             width: auto;
 


henrikhermansen:
To assure this fix is as intended, I've supplied screenshots of how the buttons will behave now.

Enough width for all buttons:
image

Not enough width for all buttons, but above breakpoint-sm:
image
Should we do something about this?

Below breakpoint-sm:
image

wkillerud:
> Not enough width for all buttons, but above breakpoint-sm. Should we do something about this?

It's only an issue in the styleguide since the available width is lower than what we normally expect. We _could_ extend the [styleguide-only CSS](https://github.com/SpareBank1/designsystem/tree/develop/src/styles) with overrides to have it break sooner.

packages/ffe-buttons/less/button-group.less
```diff
@@ -14,12 +14,12 @@
 //
 // Styleguide ffe-buttons.button-group
 .ffe-button-group {
-    display: inline-flex;
+    display: flex;
```
wkillerud, line 17: Changing direction isn't enough? This should be OK since it's designed to use 100% width on small screens, but the reason we used `inline-flex` was #173

wkillerud (approved):

packages/ffe-buttons/less/button-group.less
henrikhermansen: `inline-flex` makes the container an inline element, which in turn allows the groups themselves to stack horizontally if the buttons are narrow enough.
image

wkillerud: Tested the patch locally and it does not reintroduce #173, so we're good to go :rocket:

PR #195, opened 2018-04-20
fix(ffe-message-box-react): move to newest ffe-message-box

henrikhermansen:
This commit moves our peer dependency to the newest version of
ffe-message-box. While this is a major version change for the styling, it
has no impact on this package, as the removed class was never implemented
as a component here.


henrikhermansen:
Made an amend commit to change the commit message to a breaking change chore instead of a simple fix.

wkillerud (approved):

PR #196, opened 2018-04-20
fix(ffe-datepicker-react): fixed issues with focus, tab navigation, k…

ksawerysb1:
…eyboard input

Added new state variables to handle race conditions and tabbing focus changes between components.
Modified event handlers and fixed bugs related to focus flow.


ksawerysb1:
@selbekk added test for focus and blur of calendar component

packages/ffe-datepicker-react/src/datepicker/Datepicker.js
```diff
@@ -250,9 +253,10 @@ export default class Datepicker extends Component {
     closeCalendarSetInputFocus() {
         this.removeGlobalEventListeners();
         this.setState({
-            openOnFocus: false,
+            openOnFocus: true,
             displayDatePicker: false,
-        });
+            closeAfter: true
+        },this.dateInputRef._input.focus());
```
selbekk: `this.setState`'s second argument is a function. Pass in an arrow function to make sure `react` can call it whenever it has flushed its state changes:

```js
this.setState(
  {
    openOnFocus: true,
    displayDatePicker: false,
    closeAfter: true,
  }, 
  () => this.dateInputRef._input.focus(),
);
```

selbekk (dismissed):
This looks good to me, but I have to admit I'm a bit shaky as to what's happening here. Would be beneficial to get another person's opinion that has have more to do with this component's source code, before merging.

packages/ffe-datepicker-react/src/datepicker/Datepicker.js
ksawerysb1: Thanks, committed the change now.

selbekk (dismissed):

selbekk:
@henrikhermansen @antidecaf want to have a look?

selbekk (approved):

PR #200, opened 2018-04-23
Reintroduce the spinning shortcut button

### LINKED ISSUE #199: ShortcutButton no longer supports loading state

`ffe-buttons`/`ffe-buttons-react`

The support for a loading state seems to have been removed from `ShortcutButton` although this does not seem to be mentioned in the changelog. As a result, we're now stuck with a button that just becomes a white box when we try to set the loading state.

Breaking changes should be clearly indicated so that we can handle this when upgrading to new majors. It would be nice if this support could be re-introduced as a patch-version and then if it's not desired to support it there could be a new major release where it is explicitly removed again.

Our loading shortcut-button looks like this:

image

#### Steps to reproduce

Set `isLoading` prop on a `ShortcutButton`

ivarni:
Fixes #199.

If you don't want to support this feature in the future then please let me get a patch release and I can remove it again in a new (breaking) release with a clear indication in the changelog that it's no longer supported.

### DIFF
--- a/packages/ffe-buttons-react/src/BaseButton.js
+++ b/packages/ffe-buttons-react/src/BaseButton.js
@@ -21,7 +21,7 @@ const BaseButton = props => {
         ...rest
     } = props;
 
-    const supportsSpinner = ['action', 'primary', 'secondary'].includes(buttonType);
+    const supportsSpinner = ['action', 'primary', 'secondary', 'shortcut'].includes(buttonType);
     return (
         <Element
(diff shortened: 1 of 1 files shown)


wkillerud:
08b1979 should perhaps not be a part of this PR? Sorry for the undocumented (unintended?) breaking change.

ivarni:
Oh shoot, I accidentally a commit

wkillerud (approved):

PR #203, opened 2018-04-24
feat(ffe-tabs): new class ffe-tab-button-group

henrikhermansen:
This commit adds the fabled css class `.ffe-tab-button-group`.
Rumors of this legendary class has long been hinted by its related React
component `TabGroup`, which applied this styling to its internal `div`.

Until this very day, this class had absolutely no powers. Now this class
will give you legendary powers like plenty of padding, good margin, some
flex direction and even a media query. If you're bold, you may even apply
the `--thin` modifier to let your tab button group slip more easily
through crowded areas.

Fixes #201

PR #204, opened 2018-04-24
feat(ffe-tabs-react): add props thin and className to TabGroup

henrikhermansen:
This commit adds two new props to the `TabGroup` component.
Bool `thin` adds the `--thin` modifier to remove top and bottom padding.
String `className` adds given css classes to the tab group.

See #203 for screen shots

### DIFF
--- a/packages/ffe-tabs-react/package.json
+++ b/packages/ffe-tabs-react/package.json
@@ -33,7 +33,7 @@
   },
   "peerDependencies": {
     "@sb1/ffe-core": "^12.0.0",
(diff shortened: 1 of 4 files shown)


selbekk (dismissed):
Ser bra ut! Kjører du endringene dine gjennom prettier også?

henrikhermansen:
Oh my. I thought prettier was still prettifying my edited lines. Maybe we should re-enable that?
Anyways, I've made some new, prettified commits now.

selbekk (approved):

PR #206, opened 2018-04-24
fix(ffe-buttons-react): Restrict isLoading prop

ivarni:
BREAKING CHANGE: Removed support for `isLoading` from any button
other than ActionButton, PrimaryButton and SecondaryButton.

This commit removes support for the `isLoading` prop for all other
buttons, meaning they will not be disabled or act any differently
if given an `isLoading` prop of true or false.

If you are using `isLoading` with any other kind of button you need
to sit down with design/ux and either start using one of those
three button types or figure out another way to show a loading indicator.

### DIFF
--- a/packages/ffe-buttons-react/src/BaseButton.js
+++ b/packages/ffe-buttons-react/src/BaseButton.js
@@ -21,19 +21,21 @@ const BaseButton = props => {
         ...rest
     } = props;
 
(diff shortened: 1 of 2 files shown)


ivarni:
0.06% increased coverage! 💪

selbekk (approved):
LGTM!

wkillerud (approved):

PR #210, opened 2018-04-26
feat(ffe-radio-button-react): value can be boolean

ivarni:
Changed the prop-types of `BaseRadioButton` to accept value and
selectedValue as either boolean or string.

This means, for instance, that when using `RadioSwitch` the consumer
can easier keep state values as booleans. This can help not falling
into the "truthy-trap" when later doing branch logic based on the
value provided to a radio-switch.


selbekk (approved):
LGTM!

PR #217, opened 2018-04-30
feat(ffe-radio-button-react): value can be boolean

ivarni:
Since support for boolean value was added to `RadioSwitch` it was
now time to also support it in the `RadioButtonInputGroup` we tend
to wrap those in.


ivarni:
Dette hadde jeg sannsynligvis oppdaget før dersom `npm link` hadde fungert :|

wkillerud (approved):

PR #218, opened 2018-04-30
fix: fix intro paragraph background color

antidecaf:
Fixes a bug in which the intro paragraph in static pages in the styleguide would have the wrong background color on small screens.

image


wkillerud:
Oops :sweat_smile: Takk for fixen!

selbekk (approved):
Aha det som var riktig fiks!

PR #221, opened 2018-05-03
build(ffe-icons): create destination directory if needed

henrikhermansen:
This commit makes `ffe-icons` able to create destination directory when
building icons, if it is missing.

### DIFF
--- a/packages/ffe-icons/bin/build.js
+++ b/packages/ffe-icons/bin/build.js
@@ -4,6 +4,7 @@
 const fs = require('fs');
 const path = require('path');
 const svgstore = require('svgstore');
+const mkdirp = require('mkdirp');
 
 const ICONS_PATH = path.join(__dirname, '..', 'icons');
 
@@ -55,6 +56,8 @@ if (options.projectIcons) {
             sprite.add(iconName, fs.readFileSync(iconPath), 'utf-8');
         });
 }
+
+mkdirp.sync(options.dest);
 
 fs.writeFileSync(
     path.join(options.dest, 'ffe-icons.svg'),
--- a/packages/ffe-icons/package.json
+++ b/packages/ffe-icons/package.json
@@ -20,6 +20,7 @@
     "test": "npm run lint"
   },
   "dependencies": {
+    "mkdirp": "^0.5.1",
     "svgstore": "^2.0.3",
     "yargs": "^6.5.0"
   },


packages/ffe-icons/bin/build.js
```diff
@@ -56,6 +57,12 @@ if (options.projectIcons) {
         });
 }
 
+if (!fs.existsSync(options.dest)) {
```
wkillerud: `mkdir -p` does nothing if the directory exists, so this check can be removed.

wkillerud (dismissed):

henrikhermansen:
@wkillerud I amended the initial commit and re-added the merge from develop. Your requested changes should be done now :slightly_smiling_face:

packages/ffe-icons/bin/build.js
```diff
@@ -56,6 +57,10 @@ if (options.projectIcons) {
         });
 }
 
+mkdirp.sync(options.dest, error => {
+    if (error) { console.log(error); }
+});
```
selbekk: As far as I can see from [the docs](https://github.com/substack/node-mkdirp#mkdirpsyncdir-opts), `mkdirp` doesn't accept a function as a second argument. Actually, [looking at the source code](https://github.com/substack/node-mkdirp/blob/master/index.js#L56-L58), it looks like if the second argument isn't an object, it sets the second argument as the `mode` argument - which will potentially bork your access rights.

I suggest that you either just drop the second argument:

```js
mkdirp.sync(options.dest);
```
, or do the same, but add error handling (not sure how real of an issue that will ever be) and a proper error message (and error code) to describe what happened :)

selbekk (changes requested):

packages/ffe-icons/bin/build.js
ivarni: I was actually puzzled by why a synchronous function would accept a callback but just figured the API was wierd :)

henrikhermansen: Hmm indeed, so it seems.
I actually took this from [a different part of the docs](https://github.com/substack/node-mkdirp#example). Maybe I should file an issue about his example :stuck_out_tongue: 
I'll fix this in a bit!

ivarni: Yeah, that's the async version. It makes sense for that to take a callback.

selbekk: @henrikhermansen that part's correct though - if you want to use the asynchronous version, you need to pass a function as a callback. The sync version should probably come with its own example in the docs.

henrikhermansen: You guys are right! I'm being a total noob right now :man_facepalming:

henrikhermansen: @selbekk I simply removed the callback as second argument.

selbekk (approved):

Issue #180, opened 2018-04-17
ffe-accordion animerer pila over Z-aksen


wkillerud:
Under [animasjoner](https://design.sparebank1.no/visuell-identitet.html#visuell-identitet_animasjoner) beskriver vi at vi animerer "bevegelse i X- og Y-aksen - ikke Z (3D)". 

Vi bryter med det prinsippet i [ffe-accordion](https://design.sparebank1.no/styleguidist/index.html#accordion-1) :male_detective:  Vi bør endre animasjonen så den følger guidelines som er definert under _Visuell identitet_.

benmatthews:
The css is actually using `rotateX(180deg)` which rotates around the x axis.  I guess a simple solution to this is to just use `rotate(180deg)` but then the GPU is not going to be used... so maybe the right solution is to use `rotateZ(180deg)`

ivarni:
I'm fairly sure we can live without GPU acceleration for this one :)

adidick-zz:
I think it's safe to stick to the X-Y rules for the sake of consistency here.
I personally think the 180 spin is a bit weird visually in regards to the way the field actually expands (rotate ≠ expand) so that's something we could look into in regards to how all of our transitional animations work. It would be nice to visually support the action.

wkillerud:
Solved in #225

PR #225, opened 2018-05-04
fix(ffe-accordion): change the rotation axis for the accordian chevron animation.

benmatthews:

### DIFF
--- a/packages/ffe-accordion/less/ffe-accordion.less
+++ b/packages/ffe-accordion/less/ffe-accordion.less
@@ -45,7 +45,7 @@
     }
 
     &--open {
-        transform: rotateX(180deg);
+        transform: rotateZ(180deg);
         fill: @ffe-blue-royal;
     }
 


selbekk (commented):
I think the current implementation is nicer to look at to be honest. Where is this change coming from?

This is the current behavior: 
https://design.sparebank1.no/styleguidist/index.html#!/Accordion

benmatthews:
This change is coming from this [issue 180](https://github.com/SpareBank1/designsystem/issues/180) and there is a section in the design guide on [bevegelse i X- og Y-asken](https://design.sparebank1.no/visuell-identitet.html#visuell-identitet_animasjoner).  Maybe the correct solution is to remove the animation?

selbekk:
Aha! I think the current approach look good, but if there are arguments that we shouldn't animate stuff across the Z-axis, that's fine by me.

benmatthews:
Well, the alternative is to just not animate the transition... I can check with Adi.

wkillerud (approved):
LGTM

wkillerud (approved):
LGTM

PR #226, opened 2018-05-07
Add top padding to grid col

eirikv:
Gi mulighet for litt pusterom når bakgrunnsfarge er satt på kolonnen.

### DIFF
--- a/packages/ffe-grid/less/ffe-grid.less
+++ b/packages/ffe-grid/less/ffe-grid.less
@@ -304,6 +304,14 @@
     .ffe-grid--condensed &--no-bottom-padding {
         padding-bottom: 0;
     }
+
+    &--top-padding {
+        padding-top: @ffe-grid-gutter-sm;
+
+        @media only screen and (min-width: @breakpoint-lg) {
+            padding-top: @ffe-grid-gutter-lg;
+        }
+    }
 }
 
 .create-column(sm, 12);
--- a/packages/ffe-grid-react/src/GridCol.js
+++ b/packages/ffe-grid-react/src/GridCol.js
@@ -34,6 +34,7 @@ const MODIFIER_LIST = [
     'start',
     'centerText',
     'top',
+    'topPadding',
 ];
 
 const sizeClasses = (size, def) => {
@@ -146,6 +147,8 @@ GridCol.propTypes = {
     start: bool,
     /** Place content elements at the top of the column */
     top: bool,
+    /** If true, column receives top padding. Useful for when you have background colored columns */
+    topPadding: bool,
     /** Size modifiers for small screen sizes */
     sm: oneOfType([
         /** Number of columns, 0-12 */
--- a/packages/ffe-grid-react/src/GridCol.spec.js
+++ b/packages/ffe-grid-react/src/GridCol.spec.js
@@ -89,6 +89,12 @@ describe('GridCol', () => {
         expect(el.hasClass('ffe-grid__col--top')).toBe(true);
     });
 
+    it('sets the top-padding modifier', () => {
+        const el = renderShallow({ topPadding: true });
+
+        expect(el.hasClass('ffe-grid__col--top-padding')).toBe(true);
+    });
+
     it('sets background color class if valid', () => {
         backgroundColors.forEach(background => {
             const el = renderShallow({ background });
--- a/packages/ffe-grid/README.md
+++ b/packages/ffe-grid/README.md
@@ -169,6 +169,10 @@ Aligns the content of a cell to the right
 
 Aligns the content of a cell to the top
 
+##### `.ffe-grid__col--top-padding`
+
+Adds top-padding to the column. Useful for when you have background colored columns
+
 ##### `.ffe-grid__col--middle`
 
 Aligns the content of a cell to the vertical middle


ivarni:
Kanskje to setninger om hva dette er til på https://design.sparebank1.no/styleguidist/index.html#!/Grid/7 ?

selbekk:
LGTM. 

My only concern here is that people could start using this for other reasons than adding background colors. This could potentially be misused slightly, and could ruin the vertical flow.

Perhaps the top padding could be applied only for the background color case? I.e. by adding a background-color-class.

ivarni:
Wait, it adds padding to just a few columns? That sounds like it could get out of hand indeed. 

Why can't we just have one version of background-coloured cols? If it's too cramped maybe every col in a row with background-colour should get padded? We're moving out of grid-land if we loose the concept of an aligned row.

eirikv:
Vi påtvinger jo ikke noe for å bruke `topPadding` for rader, og i dokumentasjonen så står det at den også er der spesifikt for bakgrunnsfarget?

ivarni:
RE: Discussion on https://github.com/SpareBank1/designsystem/pull/123 there seems to be a majority vote for solving very specific things like this in userland untill there's wider need for it.

(Not 100% sure why that wasn't also done for background colours on cols to begin with)

> Vi påtvinger jo ikke noe for å bruke topPadding for rader, og i dokumentasjonen så står det at den også er der spesifikt for bakgrunnsfarget?

When you pad an entire row then you're still getting the same padding for all content on that row meaning it's still all aligned on the same horizontal line. You've no guarantee for that when you start padding individual columns.

eirikv:
Jeg ser ihvertfall ikke noe praktisk bruk av bakgrunnsfarger i kolonner uten at det mulighet for padding. Det ser jo ikke noe bra ut når innholdet ligger klint inntil toppen. Da må det være opp til konsumentene å wrappe innholdet i en egen container som legger til margin eller padding og da kan liksågodt den wrapperen ha bakgrunnsfarge.

selbekk:
My initial point stands - scrap the `topPadding` prop, and just apply top padding to any column with a background-color. This way, it's only applied when needed 👍

selbekk:
@eirikv the background color could indeed be applied by the consumer - but then it wouldn't be bleeding out into the surrounding gutters. That's why I added it to begin with. I probably should've added the padding as well, but I either forgot, or thought that it could be solved in userland.

eirikv:
Ok, hvis det er det du vil ha. :) Men du får jo samme problemet om bare 1 kolonne i en rad skal ha bakgrunnsfarge og da bare den kolonnen får padding. Hvis det bare skal legges til padding på kolonner med bakgrunnsfarge, burde ikke da radene også ha padding som default dersom hvis de har bakgrunnsfarge?
Det virker litt smidigere om begge hadde samme syntax eller at api'et er fleksibelt.

selbekk:
That's a fair point!

ivarni:
We have collapsable rows with background-color where we can't have paddingon the entire row.

I don't think we can talk about padding an entire row and padding a cell in a column like it's the same thing...

ivarni:
Changing the defaults around shouldn't be a problem though it will be a breaking change and it's already not consistent how padding is handled since it's not default on rows but is on the grid itself.

If we're gonna break that now because of this then we should perhaps batch it up with changing the number of cols API as well since every consumer will have to go over all their grids in the first place.

To be honest, I think both changes are pretty disruptive and I'd like to see a very good reason for doing it.

packages/ffe-grid/less/ffe-grid.less
```diff
@@ -304,6 +304,14 @@
...
         padding-bottom: 0;
     }
+
+    &--top-padding {
+        padding-top: @ffe-grid-gutter-sm;
+
+        @media only screen and (min-width: @breakpoint-lg) {
+            padding-top: @ffe-grid-gutter-lg;
+        }
+    }
```
selbekk, line 314: This prop should probably use the `@ffe-condensed-grid-gutter-[sm|lg]` variable when the grid is condensed.

selbekk (changes requested):
If you make sure this adheres to the condensed grid gutters when applicable, this seems like the correct way to move ahead.

The padding and margin props for this component are in need of a normalization though, but I'm not sure what that "normal" should be.

ivarni:
From what I've seen so far, the designers doesn't seem to have a model for vertical padding at all when using the grid, so I'm not even entirely convinced it should be a part of FFE untill they do.

If anything should be the norm, I'm leaning towards having no vertical padding as a default, but again, that's a pretty big breaking change for the value it provides.

selbekk:
cc @adidick - what's the norm on vertical padding in the grid?

ivarni:
I doubt you can get a clear answer for that since there will be different scenarios where different amount of padding between content is desired. E.g. the grid can be used to horizontally align content between two different sections of a page or within a sum of factors for the price of an insurance, both which would require wildly different padding.

When I was talking about the norm for vertical padding I was referring to vertical padding on cols, rows and the grid itself in the code. I'm not even sure design *wants* to have a relation to what is a row or not, aren't they mainly concerned with aligning content? In my eyes rows are an implementation detail, along with all other vertical flow.

adidick-zz:
Plenty of good comments above. In regards to the grid there hasn't been much focus on that in the last while (or ever!). I think Julie (a designer in the Bank team) has spent a bit of time looking at a horizontal grid, spacing principles and typographic hierarchy, so there's a lot that she will hopefully address there. I agree with Ivar in that it's probably a bit premature to include anything into the DS until such principles are locked downed and standardised in the design. It's been like that for a while now, so unless there's something that's ultimately breaking for you guys I reckon we just wait for a more detailed plan.

selbekk:
Alright. 

Since this ultimately can be solved in user space - I am going to close this PR for now. I still think it's something we need, but I guess there's more discussion to be had before we get there.

PR #227, opened 2018-05-07
docs: Add markup and styling documentation

antidecaf:
This commit adds documentation of markup and styling in Less components, in order to ease the reintroduction of visual testing and use of said styles in non-React apps. The documentation follows the syntax of [KSS](https://github.com/kss-node/kss-node), which generates a styleguide into `dist/markup` upon running `npm run kss:build`.


selbekk:
Wow!

Great work @antidecaf ! Quick question - could we perhaps link to the inline SVGs instead of inlining them? They are the cause of most of the noise in the docs as of now - but I understand if we need to have them.

selbekk (approved):

wkillerud:
I'm worried that the duplication of examples between KSS and Styleguidist will end up with them getting out of sync. We don't have breaking changes in the markup that often, so maybe it's not that much of an issue I guess.

I'd love it if Styleguidist could show "View HTML" as well as "View Code", but unfortunately that's not possible ([yet](https://github.com/styleguidist/react-styleguidist/issues/968)). That wouldn't solve the visual regression test problem though...

wkillerud (approved):

antidecaf:
@selbekk Yeah, I suppose that's possible. Preferably in an upcoming pr though. This markup represents the output of the corresponding React components, so it might be necessary to make the change in those too?

@wkillerud I agree, the best option would definitely be to stick with only one set of examples. Too bad Styleguidist doesn't output HTML, but let's hope that issue results in a plugin or something.

PR #228, opened 2018-05-08
feat(ffe-tables-react): added render override functions

ksawerysb1:
added render override functions, should not be a breaking change
columns: cellRender, columnHeaderRender, columnFooterRender
table: rowRender, headerRender, footerRender


packages/ffe-tables-react/src/Table.md
```diff
@@ -126,3 +126,202 @@ const onSort = ({sortBy, descending, tableData}) => {};
...
+const classNames = require('classnames');
+const HakeIkon = require('@sb1/ffe-icons-react/lib').HakeIkon;
+const KryssIkon = require('@sb1/ffe-icons-react/lib').KryssIkon;
+
+
+const generateCheckbox = (value) => {
+    return <div style={{width: '100%', textAlign: 'center'}}>{value ?
+        <HakeIkon className={classNames({'icon-green': true, 'ffe-table__expand-icon': true})}/> :
+        <KryssIkon className={classNames({'icon-red': true, 'ffe-table__expand-icon': true})}/>}</div>;
+};
```
selbekk: Remember to run Prettier on your code!

selbekk: Also, you don't need `classNames` here?

```js
<HakeIkon className="icon-green ffe-table__expand-icon" /> // etc
```

packages/ffe-tables-react/src/TableParts/TableFooter.js
```diff
@@ -3,35 +3,75 @@ import PropTypes from 'prop-types';
 import TableCell from './TableCell';
 import classNames from 'classnames';
 
-const TableFooter = ({ columns }) => {
+const TableFooter = ({ columns, footerRender, dataWindow }) => {
```
selbekk, line 6: So I'm wondering - could this be implemented by basically changing TableFooter, TableHeader and TableRow implementations to:

```js
const TableRow = (props) => {
  const cellsContent = columns.map(/* current impl */);
  const trProps = { className: 'ffe-table__row' };
  if (props.renderer) {
    return renderer({ children: cellsContent, className: 'ffe-table__row' });
  }
  return <tr>/* currentImpl */</tr>
};
```

Basically - if a renderer function is passed, call it with the rendered children (if any). if not, just use the original implementation!

This way, we can get away with much less changes, less repetition and hopefully less bugs.

ksawerysb1: This implementation allows easy reuse of already rendered default components by providing them as a parameter to the render functions, if we use children, how do would we mix overrides and original content? In the most vanilla override we just need a "<tr {...trprops}>{rowcontent}</tr>" and can just add a onClick or a class to the row.

packages/ffe-tables-react/src/Table.md
ksawerysb1: I always run prettier! :) But it seems to avoid the .md files, have to just reformat outside of the file I guess.

Yes the class names are maybe too much of an example, I will remove them.

packages/ffe-tables-react/src/TableParts/TableFooter.js
selbekk: The `renderer` prop in the example I posted above, could look like this:

```js
rowRenderer: rowProps => (
  <tr onClick={myClickHandler} {...rowProps} />
),
```

This would pass the default rendered cells as `children`. If you want to override a class name for example, it would look like this:

```js
rowRenderer: rowProps => (
  <tr 
    onClick={myClickHandler} 
    {...rowProps} 
    className={classNames(rowProps.className, 'custom-class-name')} 
  />
),
```

Note that `children` is just a regular prop as far as React knows (well, almost), so these two are equivalent:

```js
const PageHeading = props => <h1>{props.children}</h1>;
const PageHeading = props => <h1 children={props.children} />;
```

ksawerysb1: I think we discussed children overrides as a next major version and having render functions as a non-breaking change. I think this implementation gives us a lot of flexibility now, and maybe we can do a full revamp for a new major release? Children might be better for visibility of the structure inside the table, but the render functions require less boilerplate code I think.

packages/ffe-tables-react/src/Table.md
ksawerysb1: @selbekk prettiefied and cleaned up

packages/ffe-tables-react/src/TableParts/TableFooter.js
henrikhermansen: To be honest I'm having a bit of trouble fully wrapping my mind around all this. I can't fully picture the big difference in use or implementation between the PR and what is suggested by @selbekk . Both seems to be a way of passing a function, which will be ran by the table(row/col) components.
Why would optional render prop be a breaking change?

While this PR certainly is a big one, it seems to me like an acceptable addition as it is now. This whole package is very complex, and contains many little features, each in use by only a few. This seems to be yet another one of those, but at least it will meet the needs of a consumer of this package.

This PR, and this discussion, is clearly an indication of how dire the need is to rewrite this whole monstrosity :stuck_out_tongue:

ivarni: I'm thinking (in the context of a rewrite, not this PR) that this package should just provide building-blocks like `Row` and `Cell` and leave the rest to live in userland.

ksawerysb1: @henrikhermansen I think compressing all the props into one might be confusing for the end user so if the user has a list of boilerplate elements that has to be placed into for example <tr> they know that the trProps goes to <tr {...trProps}>, having it in one variable might not be that obvious that it should be placed there maybe? 
But we can squash everything into one prop, if there is a sense that this is cleaner for the end user. 
I'll take a look at the default children prop. It will save us some boilerplate, I just checked that it has the right priority.

ksawerysb1: With integrated trProps.children, and changing of the order of parameters to:
rowRender={(trprops, props, index)...
we can go down to a minimal boilerplate of:
rowRender={(trprops) => <tr {...trprops}/>}
I would still leave the separate props and index values so it's possible to use the data without overbloating the dom element.

ksawerysb1: @selbekk @henrikhermansen pushed changes for passing of elements via props.children

selbekk (approved):
Looks good to me!

PR #234, opened 2018-05-14
fix(ffe-account-selector-react): Properly aligns reset and expand button

TorgeirHeim:
screenshot-1


wkillerud:
Is this fix related to an issue? Please tag it if so :) 
Could you attach a screenshot of the end result?

packages/ffe-account-selector-react/less/base-selector.less
```diff
@@ -4,25 +4,25 @@
...
-        padding-right: 57px;
+        padding-right: 63px;
         &::-ms-clear {
             display: none;
         }
     }
 
     &__expand-button {
-        height: 23px;
-        width : 23px;
```
wkillerud, line 15: I'm a bit wary of removing `height` and `width` - could you double-check that the component still looks good in IE?  ([free ready-made VMs from Microsoft with different IE versions can be downloaded here if you need them](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/))

wkillerud (changes requested):
The changes are fine if the component still looks OK in IE, but please double-check :+1:

packages/ffe-account-selector-react/less/base-selector.less
TorgeirHeim: Verified in IE

wkillerud (approved):

PR #235, opened 2018-05-14
ffe-webfonts: License text survives minification. Also, minify CSS on design.sparebank1.no

wkillerud:
1. Minifies the CSS we produce for the styleguide using `webpack` and `lessc`.
2. Modifies the license comment so it survives minification. This way consumers don't have to copy the license text.


ivarni (commented):
LGTM

PR #239, opened 2018-05-18
feat(ffe-checkbox-react): allow children as a function

### LINKED ISSUE #158: Unable to stop onClick event from bubbling from Checkbox

henrikhermansen:
I haven't quite given up on resolving #158 with changes to the Designsystem.
I'm now taking a more general approach to the problem, with a suggested solution that could potentially be helpful for other issues as well.

This commit allows the children of Checkbox to be a function. This will
allow consumers to further customize their label node. The function is
passed one argument which is an object with props. These props should be
spread on the rendered label.

Resolves #158
Closes #162

### DIFF
--- a/packages/ffe-checkbox-react/src/Checkbox.js
+++ b/packages/ffe-checkbox-react/src/Checkbox.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { bool, node, string } from 'prop-types';
+import { bool, node, string, func, oneOfType } from 'prop-types';
 import { v4 as hash } from 'uuid';
 import classNames from 'classnames';
 
@@ -7,6 +7,14 @@ export default function CheckBox(props) {
     const { children, inline, invalid, label, noMargins, ...rest } = props;
 
     const id = props.id || `checkbox-${hash()}`;
+    const labelProps = {
+        className: classNames({
+            'ffe-checkbox': true,
+            'ffe-checkbox--inline': inline,
+            'ffe-checkbox--no-margin': noMargins,
+        }),
+        htmlFor: id,
+    };
 
     return (
         <span>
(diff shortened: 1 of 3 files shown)


packages/ffe-checkbox-react/src/Checkbox.js
```diff
@@ -17,16 +25,16 @@ export default function CheckBox(props) {
...
-                htmlFor={id}
-            >
-                {label || children}
-            </label>
+            {
+                typeof children === 'function'
+                    ? children(labelProps)
+                    : (
+// eslint-disable-next-line jsx-a11y/label-has-for
+                        <label {...labelProps}>
```
henrikhermansen, line 33: Optionally I could explicitly apply each of the labelProps to the label to avoid the eslint-disable-next-line comment.
Any thoughts?

packages/ffe-checkbox-react/src/Checkbox.spec.js
```diff
@@ -106,4 +106,17 @@ describe('<Checkbox />', () => {
         );
         expect(wrapper.find('input').prop('tabIndex')).toBe(-1);
     });
+
+    it('should render children if it is a function', () => {
+        const wrapper = shallow(
+            <Checkbox>
+                { labelProps => (
+// eslint-disable-next-line jsx-a11y/label-has-for
+                    <label {...labelProps}>Hello world</label>
```
henrikhermansen, line 115: Again, by explicitly applying each label prop to the label I could avoid the eslint disable comment.

packages/ffe-checkbox-react/src/Checkbox.js
ivarni: I think it's fine to just use the inline comment here, it clearly indicates why we're doing it.

ivarni (approved):
I don't really see any issues with this approach

selbekk:
Looks good to me! 👍

PR #242, opened 2018-05-22
fix(ffe-form-react): Make classnames a dependency

### LINKED ISSUE #232: classnames is wrongly declared as a devDependency in ffe-form-react

In the `package.json` of ffe-form-react, `classnames` is described as a devDependency. It is however imported and used directly in components, so the build will fail if it's not in place.

I think this dependency should be changed either to a normal dependency (for bundling) or to a peer dependency (to flag to the user that the lib should be provided by the including project).

BR, Preben Ludviksen, Team Finansiering

selbekk:
This commit makes `classnames` a direct dependency, and not a
dev dependency like it was.

Fixes #232

### DIFF
--- a/packages/ffe-form-react/package.json
+++ b/packages/ffe-form-react/package.json
@@ -23,11 +23,11 @@
     "setupTestFrameworkScriptFile": "../../test-setup.js"
   },
   "dependencies": {
+    "classnames": "^2.2.5",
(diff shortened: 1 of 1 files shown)


henrikhermansen (approved):
Nice, hit these little issues right outta the ball park!

antidecaf (approved):

PR #245, opened 2018-05-22
Hidden label naked checkbox

ivarni:
This does not have a corresponding issue for reasons of personal lazyness but mainly, I would like to render (and position) a checkbox without any visible label without having to work around the padding/margin that the label would have had if it was there. Which it isn't. But the padding is. Which I wantn't.

### DIFF
--- a/packages/ffe-checkbox-react/src/Checkbox.js
+++ b/packages/ffe-checkbox-react/src/Checkbox.js
@@ -1,10 +1,18 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import { bool, node, string, func, oneOfType } from 'prop-types';
 import { v4 as hash } from 'uuid';
 import classNames from 'classnames';
 
 export default function CheckBox(props) {
-    const { children, inline, invalid, label, noMargins, ...rest } = props;
+    const {
+        children,
+        hiddenLabel,
+        inline,
+        invalid,
+        label,
+        noMargins,
+        ...rest
+    } = props;
 
     const id = props.id || `checkbox-${hash()}`;
     const labelProps = {
@@ -12,12 +20,14 @@ export default function CheckBox(props) {
             'ffe-checkbox': true,
             'ffe-checkbox--inline': inline,
             'ffe-checkbox--no-margin': noMargins,
+            'ffe-checkbox--hidden-label': hiddenLabel,
         }),
         htmlFor: id,
     };
+    const LabelElement = hiddenLabel ? 'span' : 'label';
 
     return (
-        <span>
+        <Fragment>
             <input
                 className="ffe-hidden-checkbox"
                 id={id}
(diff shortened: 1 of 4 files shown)


packages/ffe-checkbox-react/src/Checkbox.js
```diff
@@ -1,23 +1,33 @@
...
+        ...rest
+    } = props;
 
     const id = props.id || `checkbox-${hash()}`;
     const labelProps = {
         className: classNames({
             'ffe-checkbox': true,
             'ffe-checkbox--inline': inline,
             'ffe-checkbox--no-margin': noMargins,
+            'ffe-checkbox--hidden-label': hiddenLabel,
```
wkillerud, line 23: Is it too magic if this class and `hiddenLabel` value is derived from the fact that `aria-label !== undefined`?

ivarni: What if someone wants an image as a label?

ivarni: I suppose you could argue that `alt` on the image would work but I would personally still prefer this to not be magic.

selbekk: What about `--no-label`, to keep with the `--no-margin` modifier? Slightly different use case of course, but it's nice to have similar naming.

selbekk (dismissed):
LGTM!

packages/ffe-checkbox-react/src/Checkbox.js
ivarni: I can change it to that if that's what you prefer, but I didn't want to encourage anyone to have a checkbox with no label what-so-ever so I went with hidden and added the bit about `aria-label`.

wkillerud: The current name is fine by me - like Ivar says we don't want to encourage not using a label, even by name :)

packages/ffe-checkbox-react/src/Checkbox.js
```diff
@@ -1,23 +1,33 @@
...
+import React, { Fragment } from 'react';
 import { bool, node, string, func, oneOfType } from 'prop-types';
 import { v4 as hash } from 'uuid';
 import classNames from 'classnames';
 
 export default function CheckBox(props) {
-    const { children, inline, invalid, label, noMargins, ...rest } = props;
+    const {
+        children,
+        hiddenLabel,
```
wkillerud, line 9: > `9:9  error  'hiddenLabel' is missing in props validation  react/prop-types`

wkillerud (dismissed):
I'm happy once Travis is happy :)

packages/ffe-checkbox-react/src/Checkbox.js
selbekk: Yeah I don't feel strongly either way, Go with `hidden`

packages/ffe-checkbox-react/src/Checkbox.js
ivarni: We don't lint on precommit? 😨

ivarni: Fixed naow

selbekk (approved):

PR #247, opened 2018-05-23
Expose vars as js

### LINKED ISSUE #246: Export colors and breakpoints as JS variables!
ffe-core

#### Describe your feature request
I want to start using CSS-in-JS in one of our apps, and I would love to have the LESS variables in ffe-core available in JavaScript as well.

I'm suggesting writing a script that would extract variables from the ffe-core LESS files, and expose them as named exports of the `ffe-core` package.

selbekk:
This pull request fixes #246 . Please refer to this issue for discussion on whether or not this is something we want to expose, while discussing the technical details here :slightly_smiling_face:

### DIFF
--- a/packages/ffe-core/bin/build.js
+++ b/packages/ffe-core/bin/build.js
@@ -1,0 +1,62 @@
+const fs = require('fs');
+const path = require('path');
+const Case = require('case');
+const mkdirp = require('mkdirp');
+
+// These are the files we want to convert to JavaScript
+const FILES_WITH_VARIABLES = ['colors', 'breakpoints'];
+
+// First, create the lib directory if it doesn't exist
+mkdirp.sync(path.resolve(__dirname, '..', 'lib'));
+
+// Then, let's process each file!
+FILES_WITH_VARIABLES.forEach(filename => {
+    const lessContent = fs.readFileSync(
+        path.resolve(__dirname, '..', 'less', `${filename}.less`),
+        'utf8'
+    );
+
+    // This is done in a ridiculously step by step fashion to increase
+    // readability. Basically, this changes a LESS file into a JavaScript object
+    const variables = lessContent.split('\n')
+        .map(line => line.trim())
+        .filter(line => line.startsWith('@') && !line.startsWith('@import'))
+        .map(line => line.replace('@ffe-', ''))
+        .map(line => line.replace('@', ''))
+        .map(line => line.split(': '))
+        .reduce((allVars, [key, value]) => ({
+            ...allVars,
+            [Case.camel(key)]: value.substring(0, value.indexOf(';')),
+        }), {});
+
+    // Create the output string, complete with a header explaining where it came
+    // from and how to change it.
+    const jsContent = `// DO NOT MODIFY!
+// This file is created automatically by ffe-core/bin/build.js, and is based on
+// the ffe-core/less/${filename}.less file. Change the value there instead.
+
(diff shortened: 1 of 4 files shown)


packages/ffe-core/less/colors.less
```diff
@@ -5,7 +5,7 @@
 @ffe-blue-sky: #7FC6E8;
 @ffe-blue-pale: #DFF1F9; // Info-boxes, etc.
 @ffe-blue-ice: #EFF8FC; // Expandable table rows
-@ffe-blue-focus: fade(#44C0FF, 50%); // Focus on buttons and controls
+@ffe-blue-focus: rgba(68, 192, 255, 0.5);; // Focus on buttons and controls
```
antidecaf: `;;`

Oops

selbekk: ah. Sorry!

selbekk: Btw - this value is taken from running the less compiler and copying its output (that's probably why the extra `;` came along). This way, we know it's the exact same result!

packages/ffe-core/bin/build.js
```diff
@@ -0,0 +1,62 @@
...
+    const lessContent = fs.readFileSync(
+        path.resolve(__dirname, '..', 'less', `${filename}.less`),
+        'utf8'
+    );
+
+    // This is done in a ridiculously step by step fashion to increase
+    // readability. Basically, this changes a LESS file into a JavaScript object
+    const variables = lessContent.split('\n')
+        .map(line => line.trim())
+        .filter(line => line.startsWith('@'))
```
wkillerud: May want to exclude `@import` in this filter.

packages/ffe-core/bin/build.js
```diff
@@ -0,0 +1,62 @@
...
+`;
+
+    // Write the formatted string to its own file
+    fs.writeFileSync(
+        path.resolve(__dirname, '..', 'lib', `${filename}.js`),
+        jsContent,
+    );
+});
+
+// Finally, let's create an index file that exports all variables
```
wkillerud, line 48: I'm a bit torn here. It would be safer (with regards to name collisions) to not have a shared index and import from `ffe-icons/colors` like we did in older versions of `icons-react`.

On the other hand, chances are low that we'll ever get a naming collision and I know we just refactored `ffe-icons-react` to use a single shared `index.js`. On the other _other_ hand, there we have a single folder and unique file names that get translated to exports. Here we have no such guarantee.

Thoughts?

wkillerud (changes requested):
The `@import` exclude should probably be added before merging

packages/ffe-core/bin/build.js
selbekk: Good point. The naming collision thing is already handled by the fact that this is a mirror of LESS variables, and LESS-variables are global. If we would ever get a naming collision, it would impact our LESS as well, which would probably be caught at some point (compile time? idk). 

Exporting variables with the same name is a bad practice we should try to avoid anyhow.

Due to limitations in npm, the imports would have to be `ffe-core/lib/colors` (can be hacked around, but it's not a nice solution). Most npm packages only have a single top level place to import from - I like that pattern.

wkillerud: > LESS-variables are global

Good point! Then it's a non-issue. Carry on :)

wkillerud (approved):

antidecaf (approved):

PR #249, opened 2018-05-23
feat(ffe-core-react): Expose named exports

### LINKED ISSUE #248: Missing index-file for ffe-core-react
ffe-core-react

Currently, I have to import components from the lib folder like this:

```js
import Paragraph from '@sb1/ffe-core-react/lib/typography/Paragraph';
import DividerLine from '@sb1/ffe-core-react/lib/typography/DividerLine';
```

I would love an index file that exported all of these components as named components, so I could use it like so:

```js
import { Paragraph, DividerLine } from '@sb1/ffe-core-react';
```

Not sure when the original index disappeared - perhaps it was never added (by myself)?

selbekk:
This PR creates an index file which exposes all components as named
exports from the top level `@sb1/ffe-core-react` import.

Previously, the user would have to dig into the `node_modules` to figure
out that the components were located under `lib/typography/..`. Now, the
consumer can import components like this:

```js
import { Paragraph, Heading2, Heading3 } from '@sb1/ffe-core-react';
```

Fixes #248

### DIFF
--- a/packages/ffe-core-react/src/index.js
+++ b/packages/ffe-core-react/src/index.js
@@ -1,0 +1,8 @@
+export { default as DividerLine } from './typography/DividerLine';
+export { default as EmphasizedText } from './typography/EmphasizedText';
+export * from './typography/Heading';
+export { default as MicroText } from './typography/MicroText';
+export { default as Paragraph } from './typography/Paragraph';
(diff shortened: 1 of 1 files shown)


henrikhermansen (approved):
Splendid!

antidecaf (approved):

PR #257, opened 2018-05-25
256 background colours

ivarni:


benmatthews (dismissed):

selbekk (dismissed):

packages/ffe-grid-react/src/Grid.md
```diff
@@ -91,6 +95,11 @@ const backgroundColors = [
     'grey-cloud',
     'purple-magenta',
     'sand',
+    'sand',
```
antidecaf: `'sand'` seems to have been repeated in two lines.

ivarni: It's because there's a lot of it :D

antidecaf (approved):

packages/ffe-grid-react/src/Grid.md
wkillerud: > It's coarse and rough and irritating and it gets everywhere.

wkillerud (approved):

Issue #244, opened 2018-05-22
Accessibility error message in ffe-account-selector-react


eksb1:
#### Which package is this issue related to?
ffe-account-selector-react
#### Describe your issue (screenshots welcome!)
Accessibility error message: A button is empty or has no value text for close button and expand-list button in account selector.

#### What is the expected behavior? 
These buttons/symbols are not available with Tab, so area value could be disabled. 

#### Steps to reproduce
could be reproduced by using Wave-extension in Chrome

selbekk:
#258 fixed this.

PR #258, opened 2018-05-25
fix(ffe-account-selector-react): fixed accessibility errors

eksb1:

### DIFF
--- a/packages/ffe-account-selector-react/src/components/account-selector/AccountSelector.js
+++ b/packages/ffe-account-selector-react/src/components/account-selector/AccountSelector.js
@@ -96,6 +96,7 @@ class AccountSelector extends Component {
                     {...this.props}
                     onSelect={this.onAccountSelect}
                     onChange={this.onInputChange}
+                    locale={locale}
                 />
                 {selectedAccount && (
                     <AccountDetails
--- a/packages/ffe-account-selector-react/src/subcomponents/input-field/InputField.js
+++ b/packages/ffe-account-selector-react/src/subcomponents/input-field/InputField.js
@@ -101,6 +101,7 @@ class Input extends Component {
                         onMouseDown={this.onReset}
                         tabIndex={-1}
                         type="button"
+                        aria-label="reset"
                     >
                         <KryssIkon className="ffe-base-selector__reset-button-icon"/>
                     </button>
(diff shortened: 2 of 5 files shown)


eksb1:
How do I link  pull request to issue?

wkillerud:
You can link to an issue (or PR) by typing `#` and the issue (or PR) number in a comment. Github will create a link. So `#244` becomes #244 :rocket:

packages/ffe-account-selector-react/src/subcomponents/input-field/InputField.js
```diff
@@ -101,6 +101,7 @@ class Input extends Component {
                         onMouseDown={this.onReset}
                         tabIndex={-1}
                         type="button"
+                        aria-label="reset"
```
wkillerud: This text will be read out loud by screen readers and should follow the same language settings as the rest of the component. If you look in the [i18n](https://github.com/SpareBank1/designsystem/tree/develop/packages/ffe-account-selector-react/src/i18n) folder you'll find the languages and texts that are being used in the component. `RESET_SEARCH` looks promising to use here, and the label for the expand/collapse button could perhaps use `ACCOUNTSLIST_OPEN` and `ACCOUNTSLIST_CLOSE` depending on state? 

The [AccountNoMatch](https://github.com/SpareBank1/designsystem/blob/36dc3a255a300b3639b727ccdef8729c24a6a036/packages/ffe-account-selector-react/src/subcomponents/account/AccountNoMatch.js) component has an example on how to use the language file. You need to know the current `locale` though, which `Input` doesn't right now. From what I can see `Input` is only used from `BaseSelector`. `BaseSelector` gets the `locale` prop from `AccountSelector` and `AccountSelectorMulti`, it's just not used/listed in `propTypes`. So if you send the `locale` prop from `BaseSelector` to `Input` and use the language file this should be good to go :)

wkillerud (changes requested):
The labels should use texts from the language files and follow the `locale` prop.

eksb1:
#244

eksb1:
@wkillerud could you please review again?

selbekk (approved):
LGTM! Great work!

packages/ffe-account-selector-react/src/components/account-selector/AccountSelector.js
```diff
@@ -96,6 +96,7 @@ class AccountSelector extends Component {
                     {...this.props}
                     onSelect={this.onAccountSelect}
                     onChange={this.onInputChange}
+                    locale={locale}
```
wkillerud, line 99: Just for reference in case others are wondering: `AccountSelectorMulti` already passed on `locale`

wkillerud (approved):
LGTM, thanks for making the changes! :)

PR #261, opened 2018-05-29
fix(ffe-checkbox-react): Fix hidden label boxes

ivarni:
The first iteration of the `<Checkbox>` component with the `hiddenLabel`
modifier replaced the `<label>` element with a `<span>`. This of course
breaks the link between the label and the input element meaning no click
event was ever going to make it to it's intended destination.


ivarni:
So.. is that coverage check a tad bit strict?

wkillerud:
Ah, that one was subtle. Thanks for the fix.
Regarding the coverage requirement, seems like it's too strict in this case. The coverage for the component is still 100% :/ OTOH I'd rather not reduce it to avoid "death by a thousand cuts" if coverage for a component does in fact go down. An admin can override it - I'll just do that in this case.

wkillerud (approved):

Issue #260, opened 2018-05-28
New type of card(?) for overview and landing pages


henrikhermansen:
#### Which package is this issue related to?
ffe-cards and ffe-cards-react, I suppose

#### Describe your issue (screenshots welcome!)
We have a new type of card (or whatever it is to be named) which is already in use on banking overview beta, and is soon to come to various banking landing pages.
I suppose insurance will use this as well, but I don't know.

There are two types of cards.
The bigger and badder one:
![image](https://user-images.githubusercontent.com/435037/40616776-727052b0-628c-11e8-9133-2c20724329da.png)
The leaner and meaner one:
![image](https://user-images.githubusercontent.com/435037/40616780-746b0a06-628c-11e8-8e6b-0a6840cdc3e3.png)

I suppose the latter could be viewed as a condensed version of the first, but they clearly have distinct uses. Condensing here is not something that would be done for smaller size screens or anything. They should both be responsive in each their way.

Is this going to be categorized as a card?
If so, what should the card name be?
If not, what is it really?

While these "cards" are still a design WIP, they're already implemented in one place, and will surely be implemented in many more. I say surely, because this is now a well processed idea from the overview team (I believe).
Thus it makes sense to add them here, so we can all work together on making the best solution, rather than implementing many different versions before we try to merge them all together.

ivarni:
Are they going to be used around grids and is (e.g.) the text then going to be expected to line up with other grid column content? Same question for the lean one re the condensed grid.

To me this *feels* like it's not really much more than a border-shadow around a box, especially if we could use the grid for laying out the internal parts of it (in which case it is tempting to just suggest adding support for border-shadow on grid-cols).

If it can't be implemented with grid-elements I'd suggest making the implementation as flexible as possible, maybe just a border around a box with an icon on the left and whatever the consumer wants to put there on the right.

henrikhermansen:
For now my idea is to put these as components inside grid columns.

I'm not sure how specific or flexible we want these to be.
I think this is something we should have design involved in.

I've been working on some code, and so far it's not as flexible as you describe it. This is because I believe there are certain design limitations/requirements to these cards that needs to be taken into account.
The component is very similar to the ProductCard. This is because we need to control the content a bit, due to said design considerations.
However, I've tried to keep the css simple. So far it's 89 lines, which is ways better than our ffe-link-card which is 237 lines (not counting comments).

I'll put up a PR tomorrow, so you can see for yourself.
[...]

PR #262, opened 2018-05-29
Feature/some kind of card

henrikhermansen:
**This PR should not be merged**
For now, it is here for discussing styling and component API.

This relates to #260

### DIFF
--- a/packages/ffe-cards-react/src/NewCard.js
+++ b/packages/ffe-cards-react/src/NewCard.js
@@ -1,0 +1,57 @@
+import React from 'react';
+import classNames from 'classnames';
+import { node, string, bool } from 'prop-types';
+
+import GenericCard from './GenericCard';
+
+const NewCard = props => {
+    const {
+        className,
+        details,
+        heading,
+        icon,
+        condensed,
+        greyCharcoal,
+        ...rest
+    } = props;
+
+    return (
+        <GenericCard
+            className={ classNames(
+                'ffe-new-card',
+                { 'ffe-new-card--condensed': condensed },
+                { 'ffe-new-card--grey-charcoal': greyCharcoal },
+                className
+            ) }
+            { ...rest }
+        >
+            {icon &&
+            React.cloneElement(icon, {
+                className: classNames(
+                    'ffe-new-card__icon',
+                    icon.props.className
+                ),
+            })}
+            <div className="ffe-new-card__body">
+                {heading && <p className="ffe-new-card__heading">{heading}</p>}
+                {details && <p className="ffe-new-card__details">{details}</p>}
+            </div>
+        </GenericCard>
+    );
+};
+
+NewCard.propTypes = {
+    className: string,
+    /** Text shown at the top */
+    heading: string,
+    /** Bottom text */
+    details: node,
+    /** A rendered icon */
+    icon: node,
+    /** Smaller icon and less space */
+    condensed: bool,
+    /** Icon and text will all be ffe-grey-charcoal */
+    greyCharcoal: bool,
+};
+
+export default NewCard;
--- a/packages/ffe-all.less
+++ b/packages/ffe-all.less
@@ -5,6 +5,7 @@
 @import './ffe-buttons/less/buttons';
 @import './ffe-cards/less/ffe-link-card';
 @import './ffe-cards/less/ffe-product-card';
+@import './ffe-cards/less/cards';
 @import './ffe-chart-donut-react/src/ffe-chart-donut';
 @import './ffe-context-message/less/ffe-context-message';
 @import './ffe-datepicker/less/datepicker';
--- a/packages/ffe-cards-react/src/GenericCard.js
+++ b/packages/ffe-cards-react/src/GenericCard.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import classNames from 'classnames';
+import { oneOfType, node, func, string } from 'prop-types';
+
+const GenericCard = props => {
+    const {
+        className,
+        element: Element,
+        children,
+        ...rest
+    } = props;
+
+    return (
+        <Element
+            className={ classNames(
+                'ffe-generic-card',
+                className
+            ) }
+            { ...rest }
+        >
+            {children}
+        </Element>
+    );
+};
+
+GenericCard.defaultProps = {
+    element: 'button',
+};
+
+GenericCard.propTypes = {
+    className: string,
+    children: node,
+    /** The element to render the card as. */
+    element: oneOfType([func, string]),
+};
+
+export default GenericCard;
--- a/packages/ffe-cards-react/src/GenericCard.md
+++ b/packages/ffe-cards-react/src/GenericCard.md
@@ -1,0 +1,8 @@
+Dette er basisen for alle kort. Komponeten kan brukes direkte for å få en ramme du kan fylle
+dersom du trenger et spesialtilpassert kort.
+
+```js
+<GenericCard>
+    Et helt generisk kort
+</GenericCard>
+```
--- a/packages/ffe-cards-react/src/GenericCard.spec.js
+++ b/packages/ffe-cards-react/src/GenericCard.spec.js
@@ -1,0 +1,27 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import GenericCard from './GenericCard';
+
+const getWrapper = props => shallow(
+    <GenericCard { ...props }>
+        Hello world
+    </GenericCard>
+);
+
+describe('GenericCard', () => {
+    it('should render a button element with correct class and children', () => {
+        const wrapper = getWrapper();
+
+        expect(wrapper.find('button').exists()).toBe(true);
+        expect(wrapper.hasClass('ffe-generic-card')).toBe(true);
+        expect(wrapper.text()).toBe('Hello world');
+    });
+
+    it('should render my custom element and custom class', () => {
+        const wrapper = getWrapper({ className: 'my-custom-class', element: 'div' });
+
+        expect(wrapper.find('div').exists()).toBe(true);
+        expect(wrapper.hasClass('my-custom-class')).toBe(true);
+    });
+});
--- a/packages/ffe-cards-react/src/NewCard.md
+++ b/packages/ffe-cards-react/src/NewCard.md
@@ -1,0 +1,33 @@
+Dette er en ny type kort som ikke har fått navn enda. Det har plass til ikon, overskrift og detaljer.
+
+```jsx
+const { GrafOppIkon } = require('../../ffe-icons-react');
+
+<NewCard
+    heading="Aksjesparekonto"
+    details="Sparing i fond og aksjer er ikke bare for proffene"
+    icon={ <GrafOppIkon /> }
+/>
+```
+
+Kortet har to modifiers, en for å gjøre det litt mindre og en for å gjøre innholdet grått
+
+```jsx
+const { SparegrisIkon, KryssSirkelIkon } = require('../../ffe-icons-react');
+
+<React.Fragment>
+    <NewCard
+        heading="Sparekonto voksen 25"
+        details="7 004,00"
+        icon={ <SparegrisIkon /> }
+        condensed={ true }
+    />
+    <NewCard
+        heading="Start ny sparing"
+        details="Det er mange måter å spare på"
+        icon={ <KryssSirkelIkon style={ { transform: 'rotate(45deg)' } } /> }
+        greyCharcoal={ true }
+        condensed={ true }
+    />
+</React.Fragment>
+```
--- a/packages/ffe-cards-react/src/NewCard.spec.js
+++ b/packages/ffe-cards-react/src/NewCard.spec.js
@@ -1,0 +1,50 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import NewCard from './NewCard';
+import SparegrisIkon from '@sb1/ffe-icons-react/lib/sparegris-ikon';
+
+const getWrapper = props => shallow(
+    <NewCard { ...props } />
+);
+
+describe('NewCard', () => {
+    it('should render correct class and contain a div with body class', () => {
+        const wrapper = getWrapper();
+
+        expect(wrapper.hasClass('ffe-new-card')).toBe(true);
+        expect(wrapper.find('div.ffe-new-card__body').exists()).toBe(true);
+    });
+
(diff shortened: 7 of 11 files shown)


packages/ffe-cards-react/src/NewCard.js
```diff
@@ -0,0 +1,62 @@
...
+        >
+            {icon &&
+            React.cloneElement(icon, {
+                className: classNames(
+                    'ffe-new-card__icon',
+                    icon.props.className
+                ),
+            })}
+            <div className="ffe-new-card__body">
+                {heading && <p className="ffe-new-card__heading">{heading}</p>}
```
henrikhermansen, line 36: Should the card heading be some kind of <h*> semantically?
If so, this needs to be decided by the consumer to make sure it's correct in their hierarchy of headings.

packages/ffe-cards-react/src/NewCard.js
```diff
@@ -0,0 +1,62 @@
...
+                {details && <p className="ffe-new-card__details">{details}</p>}
```
henrikhermansen, line 37: As it is now we have no children prop, so I don't feel like this is really react-ish.

A possible solution could be to replace `heading` and `details` and put `children` inside the `ffe-new-card__body`-div. This would make it easier for consumers to mess up the intended design, even if we say "we expect children to be a heading and a paragraph" and then we style these as they should be.

Thoughts on this predicament?

henrikhermansen:
**Some screenshots of the current design**

image
_Cards in a grid_

image
_Condensed cards in a grid_

image
_A condensed card with greyCharcoal_

image
_This grid has 3 columns. It gets a little weird, but not too bad, right before `@breakpoint-md`_

image
_With 3 columns, it's time to break it down to just 1 when it gets narrow. This is just below `@breakpoint-md`_

packages/ffe-cards/less/ffe-new-card.less
```diff
@@ -0,0 +1,107 @@
...
+    background: @ffe-white;
+    box-shadow: 0 1px 4px 0 rgba(38, 38, 38, .3);
+    display: flex;
+    flex-flow: row nowrap;
+    margin: 0 0 10px;
+    padding: 10px 20px;
+    border: 2px solid @ffe-white;
+    border-radius: 5px;
+    outline: none;
+    cursor: pointer;
```
henrikhermansen: Should all cards be clickable?
If not, I suppose we only want the cursor on clickable cards.
This is potentially something the consumer could be responsible for.

packages/ffe-cards/less/ffe-new-card.less
```diff
@@ -0,0 +1,107 @@
...
+    text-decoration: none;
+
+    &:focus {
+        box-shadow: 0 0 0 2px @ffe-blue-focus;
+    }
+
+    &:focus,
+    &:active,
+    &:hover {
+        border-color: @ffe-blue-azure;
```
henrikhermansen: Should all cards have a hover effect?
Perhaps not the ones that aren't clickable?
I assume the answer to this will be the same for focus.

packages/ffe-cards-react/src/NewCard.js
selbekk: i don't think so - it's not a heading as much as it is the text content of most cards.

packages/ffe-cards-react/src/NewCard.js
selbekk: A component doesn't need to have children - in cases like these, they typically don't (since things need to be in a certain order.

An alternative could be to use render props with what's known as compound components:

```js
<NewCard>
  {({ Heading, Details }) => (
    <Fragment>
      <Heading element="h2">My heading</Heading>
      <Details>This is some details</Details>
    </Fragment>
  )}
</NewCard>
```

, but I think that's major overkill.

Strict APIs that makes sure the consumer doesn't abuse a component's intended usecase might even lead to better DX (developer experience) in the long term, because we know what to use what component for.

packages/ffe-cards/less/ffe-new-card.less
selbekk: Most cards should probably be clickable. Therefore I think using a `<button />` as the default Element in the React implementation might be a good idea. This way, we get a lot for free! Remember to reset `appearance` if you do though.

packages/ffe-cards-react/src/NewCard.js
ivarni: You can always accept either a `string` or a `node`

packages/ffe-cards-react/src/NewCard.js
ivarni: Hard to answer that without knowing how this component will be used in 1 year's time ;)

selbekk (dismissed):
I think this looks good tbh - but why do we need a new component? Isn't this just a revamp of the original card components?

Also, if we're adding this, I would like a comprehensive list of usecases (ref your questions about clickability) and when NOT to use it.

packages/ffe-cards/less/ffe-new-card.less
ivarni: You already accept `element` as a prop, if it's supposed to be a link, perhaps the consumer can provide `"a"` as the value and the cursor should act accordingly (I'd assume).

packages/ffe-cards-react/src/NewCard.js
henrikhermansen: That's a cool technique @selbekk ! I never considered you could return components for render props.
That's true @ivarni . I just hope we can agree on some way to do it now. Nothing stands in the way of breaking that in a year's time and doing these cards in a different way.

packages/ffe-cards/less/ffe-new-card.less
henrikhermansen: I'll look a bit into this, and see how `button` and `a` will affect the box.

adidick-zz: As per design definition, ALL cards should be clickable. Anything else is just a box grouping a set of information.

packages/ffe-cards/less/ffe-new-card.less
adidick-zz: Same again, all cards are clickable so should have a hover effect.

henrikhermansen:
@selbekk I agree this component is extremely similar to `ProductCard`. The difference is the class and a couple of modifiers not available to product card.
Is there a good way to melt these components together? I don't think I could just overwrite the old one, as people might be using it the way it is now.

packages/ffe-cards/less/ffe-new-card.less
```diff
@@ -0,0 +1,107 @@
+.ffe-new-card {
```
henrikhermansen: One thought on the styling that makes out the basics of this box. Perhaps it should be extracted as a totally separate class? I know people are using custom boxes with the same basics, so it might be good to help facilitate those boxes.
The new React component here could simply apply both classes `.ffe-card-basics` and `.ffe-new-card` (or whatever  we call them).

packages/ffe-cards-react/src/NewCard.js
henrikhermansen: If it shouldn't be a heading tag anyways, I think accepting only string is fine here.

packages/ffe-cards/less/ffe-new-card.less
henrikhermansen: Thanks for clearing that up, Adi.

I see I need to add `text-align: left` if button is used as element.

Also, if element is `"a"` I get the pointer on the entire box by default, but as they should all be clickable, I think we should keep the cursor styling.

adidick-zz:
Question: 
Does the naming of a card have to reflect it's use, eg. "ProductCard", or could we simplify it to just having one card that has multiple variations, depending on what it's going to be used for and where?

For example, product card, would just be "card with icon, title and subtext" from these examples. 
image

Or its it that in the implementation the component itself "product card" needs a name, and then could potentially take in use "card with icon, title and subtext" as it's master template.

I'm working with cards now in the sketch design file so wonder whats the best way to document them here as well.

henrikhermansen:
@adidick I hope I fully understood your question, and I believe the answer is "a bit of both".

The way I see it now we could perhaps approach this in a similar way we've done with buttons.

We could have a base card class (which simply is the frame with hover and focus) and a component (which simply takes children). This will be used as a base for the other cards, and could be used when something completely custom is needed.

With this we could add specific styling and components for each type of card. Here, we should try to do some grouping. For example your two last cards are extremely similar, but they are clearly different from the cards above them. The way I see it the two last cards are the same, they are just variations on which types of they text they contain. The same goes for the three first cards. They are they same, give or take an image and a details text.
This will come down to balancing having too complex classes and components, and having too many classes and components.
The complexity of all this depends on how many variations of cards designers can come up with.. :smile: 

To better answer your questions: the naming of a card doesn't _have_ to reflect its use, but it might help to understand where, how and why it should be used. At the same time, individual cards needs a name. Like I said, we should balance complexity and diversity, so we'll have a certain number of cards, each with a certain set of variations.

Any thoughts on this @selbekk @ivarni ?

antidecaf:
I agree with this approach from a styling perspective. All cards share quite a few base properties, and the variations of internal layouts in each card probably shouldn't have to be _that_ complex (?).

```
.ffe-base-card {
    // common styles (including a default layout)
    
    &--modifier {
        // text-align right or whatever other necessary modification
    }
}
```

adidick-zz:
@henrikhermansen 
Nicely clarified. 
You've pretty much summed up what I was alluding to. Simple, structured, common sense structure that is both flexible but, as you said, helps guide both designers and developers to use the right card in the right situation. 
With that in mind it's maybe worth taking a bit of an audit of all of the cards and their uses (currently in prod and in working design sketches) and trying to find some natural groupings and uses so that we can start looking at how to best categorise the different types.
There will certainly be some cards that are used more often than others, eg. "navigation card", "product card", so once we start gathering the variations and the context in which they're used it should be relatively easy to start categorising creating the different types in a family.

henrikhermansen:
@adidick thanks :blush:
We should definately go through all the cards to get a good overview before we implement them all.

I suggest I proceed with this PR by creating a basic class and component, as well as an initial implementation of the type of card I have created. If we do this implementation well it can serve as a guideline for implementing the rest of the cards later.

When we do the broader implementation of cards, we might want to consider removing the product card and link card which exists today. Perhaps these should be marked as deprecated in this PR as well?

ivarni:
I thought the discussion in the HipChat room concluded that not every app will nessescarily get the grey background?

ivarni:
> [11:32 AM] Adrian William Dick: Felles strategi? Hva er dette?...
[11:33 AM] Adrian William Dick: Det er kanskje at flere tenker sånn men ift en ordentlig strategi er jeg usikker på.
[11:35 AM] Adrian William Dick: Det er noen designere på bank som har sett på variasjoner av sand og lys grå som er litt varmere og at vi tenker kommer til å erstatte noen av disse eksisterende farger som finnes. Men jeg er litt usikker hva status er på denne saken.

Seems premature to deprecate the old ones unless we plan to use the new ones on white backgrounds as well

henrikhermansen:
@ivarni is your question about grey background directed to my suggestion of deprecation? Or was it something else?

ivarni:
I was under the impression that the new cards won't play nice with a white background, but if that assumption is wrong and we can move to the new ones, then sure.

But then we should look at if the new ones also fit the usecases we already have for the old.

Or we could just hold off deprecating anything :)

henrikhermansen:
Hmm I'm not sure, and I haven't tried. I can have a look at how it looks tomorrow.
Are the new cards supposed to be fine on white background @adidick ?

I'm in no rush to deprecate anything, I was just asking :)

adidick-zz:
Sorry, to clarify, the new cards MUST work on white backgrounds as well, otherwise they're of no use and the last thing we want to do is have to create a custom mongrel card that works only on white. The new card design with shadow works on white but is maybe not the best it could be in terms of accessibility, so that's something we'll just have to tweak as we go along. 

What I was thinking is that we edit the existing card style (white box with grey border) to the new style (white box, no border, drop shadow) and then tweak the shadow accordingly once we've seen how it looks/works in different situations. A white box will always look/work better on a coloured background because it'll have a better contrast than if it was on a white background, so the main thing to do here is make sure that the new card with shadow is good enough on a white background.

adidick-zz:
@ivarni In regards to use cases, I think we can make a hypothesis that the new card will work on white (it should, it's really just a question of tweaking the contrast so it still looks good visually but is also accessible), and then hopefully use-cases will show up along the way to show us whether that is true or not. If not, we adjust the new design.

ivarni:
OK, fair enough. Here's a few usecases I found in one of our apps now

image
This does not seem to be supported by the new implementation, so if we're going to keep having that status banner we'll need a new design for these or the new ones need to support it.

image
These seem like we can just use the new design directly as long as it looks nice with just the heading.

image
These aren't really cards per se, but they follow a similar pattern with the bordered box approach.

image
Here's a wider variant we're also using.

These are the ones I could find that could be (or already are) classified as cards. There might be others as well.

henrikhermansen:
I've split the styling into a base class and a class for the card I've made.
Because of the size of the styling of our existing cards, I suggest we re-write them using the new base class. This should be a separate task imo.

@ivarni I see you have a nice selection of cards in your apps. I suppose it's a job for the designers to see if all of these should be commonly available cards, or if some are too specific, and should be local in your apps. For this you should be able to use the base class and component I've split out.

New commits are coming later today or tomorrow :)

ivarni:
Having access to some kind of baseclass with the border would be very nice :)

henrikhermansen:
New commits, ready for some feedback!

No need to mention I need to write tests. I know..
Edit: Coveralls is giving me a hard time over this already :roll_eyes:

henrikhermansen:
This is what the cards currently look like

image
_GenericCard doesn't have a max-width or min-height_

image
image
_NewCard does not support the field "Kortnavn" from @adidick's sketches
It does look like his and these are the same type of card,
so should I perhaps go ahead and add that field?_

ivarni (dismissed):
This LGTM, though I'd love to see a better name for this. And some tests. :trollface:

henrikhermansen:
Tests are added. Still need a proper name for `NewCard` before this can be merged.

antidecaf:
@henrikhermansen Any news on this?

henrikhermansen:
There is news! I have resumed work on this.

henrikhermansen:
I'll close the PR for now so I can make commits while I work, without notifying you all every time.

PR #263, opened 2018-05-29
fix(ffe-message-box-react): Custom icons get default style

### LINKED ISSUE #253: Icons can not be supplied as-is to ffe-message-box-react

`ffe-message-box-react`

I'm trying to provide a custom icon to [MessageBox](https://design.sparebank1.no/styleguidist/index.html#messagebox).

It seems like the CSS that's [normally applied to the default icons](https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-message-box-react/src/SuccessMessageBox.js) does not get set when providing a custom icon. I would, as a consumer, like to not have to provide these styles myself.

I was hoping for something like this:
image

I got this:
image

#### Steps to reproduce

```xml
<TipsMessage
    icon={<HandlevognIkon />}
    title="Forutsetninger for kjøp"
>
```

ivarni:
When providing a custom icon to any of the message-boxes there is
not longer any need to manually set the size of the icon in order
to make it fit properly.

Fixes #253


packages/ffe-message-box-react/src/BaseMessage.js
```diff
@@ -21,7 +26,7 @@ const BaseMessage = props => {
                     `ffe-message-box__icon--${type}`,
                 )}
             >
-                {icon}
+                {React.cloneElement(icon, { style: iconStyles, ...icon.props })}
```
selbekk, line 29: Could you add the possibility of overriding these styles in part?

```js
React.cloneElement(icon, { ...icon.props, style: { ...iconStyles, ...icon.props.style } });
```js

selbekk (approved):

packages/ffe-message-box-react/src/BaseMessage.js
ivarni: I don't follow...?

ivarni: Oh, that. I was asking about opinions on that in the issue but didn't get any so I decided to just keep it simple.

PR #265, opened 2018-05-31
feat(ffe-form-react): Pass refs for input and textarea

### LINKED ISSUE #231: Forwarding refs to <Input>-component in ffe-form-react

This is more of a technical feature request. I am working on another internal component at SB1, that adds managed formatting to an input. I want to use the `Input` from ffe-form, but need to access the DOM ref to manage selection/cursor position. I currently achieve this with ReactDOM.findDOMNode().

There are two potential ways to improve this:

1. Give Input a new property `inputRef`, that takes a callback and passes it to `ref` on the contained <input>.

2. Upgrade to React 16.3.0 and use the new forwardRef API.

If there is consensus on such a change, I could make a PR.

BR, Preben Ludviksen

selbekk:
BREAKING CHANGE: Requires React 16.4.0 and above!

This commit lets the consumer pass refs down to the `Input` and
`TextArea` components, to handle focus et al.

The only breaking change is the fact that React 16.4.0 is required.

This commit also disables two test suites, since Enzyme doesn't yet support
React's newest versions. For more information, please refer to this
issue:

https://github.com/airbnb/enzyme/pull/1513

The support sounds like they're right around the corner it's awaiting a release), but I think this PR could be reviewed (and perhaps merged) before enzyme get their stuff together.

Fixes #231.


packages/ffe-form-react/package.json
```diff
@@ -33,16 +33,16 @@
...
+    "react-dom": "^16.4.0",
     "react-motion": "^0.5.2",
     "react-test-renderer": "^16.2.0",
     "sinon": "^4.2.1"
   },
   "peerDependencies": {
     "@sb1/ffe-core": "^12.0.0",
     "@sb1/ffe-form": "^9.0.0",
-    "react": "^16.2.0"
+    "react": "^16.3.0"
```
wkillerud: This should mirror the `devDependencies` version since consumers will actually need `16.4.x`

packages/ffe-form-react/src/Input.spec.js
```diff
@@ -22,6 +25,15 @@ describe('<Input />', () => {
         expect(wrapper.hasClass('ffe-input-field')).toBe(true);
     });
 
+    it('passes refs', () => {
```
wkillerud: Maybe expand the description here with why this is needed? Future Developer should be happy with not having to search through the Git log 😉

wkillerud (changes requested):

wkillerud:
We can make an exception with regards to the code coverage (use admin privileges), but I'd like it if this could wait until next week in case Enzyme is updated. Un-skipping these tests is one of those things that is quickly postponed and/or forgotten.

henrikhermansen:
I agree it's nice to keep up with news from React, but I also agree it's better to hold it off until Enzyme is ready to keep up. Although this probably isn't breaking any code, it would be nice to actually verify it with the tests before merging.

selbekk:
I could solve this a different (and non-breaking way) by letting the user pass in an `innerRef` or `inputRef` prop instead of using the actual `ref` prop. It would (could) lead to a breaking change in the future, but it could be combined with updating the React dependency, 

What do you think - is that a better approach perhaps?

ivarni:
It seems like a lot of extra work just for some tests..

henrikhermansen:
@ivarni I'm confused about which path you're suggesting we take here :smile:
@selbekk I think that might be a better choice, for now. But I don't feel strongly about either way.

ivarni:
I'm saying I don't think we should create an API with the intention of breaking it later just over some test library that isn't updated yet.
I'd say either disable those tests and go for the API we want or, if noone is dying to get this functionality, leave the PR open untill enzyme gets updated, but then leave the issue open so we don't forgettaboutit.

packages/ffe-form-react/src/Input.spec.js
selbekk: I mean, it's a feature / quirk of React - not sure if I want to go around and explain all of those. I'll add a comment, sure, but I don't think it's needed in a general sense.

wkillerud (approved):

selbekk:
Closing this in favor of a new branch (this one got stale).

Issue #219, opened 2018-05-02
Buttons in an inline button group doesn't align text correctly


henrikhermansen:
#### Which package is this issue related to?
ffe-buttons

#### Describe your issue (screenshots welcome!)
When wrapping buttons in an inline button group styled anchor tags doesn't get vertically aligned text.
The problem is only visible if some other button in the button group has text which is wrapped to two lines.
![image](https://user-images.githubusercontent.com/435037/39510637-a74273e4-4dea-11e8-8554-6dee97790c2b.png)
The first and the last buttons are actually anchor tags. See supplied HTML.

I don't know if the same is true for other buttons than secondary, but I would assume it's at least true for all `ffe-button`s.

Note that this issue specifically applies to inline button groups. Non-inline groups works as intended.

#### What is the expected behavior? 
Text should be aligned the same whether it's an anchor or a button that's styled...

#### What is the actual behavior?
...but it's not :P 

#### Steps to reproduce
```
<div class="ffe-button-group ffe-button-group--inline">
    <a class="ffe-button ffe-button--secondary ffe-button--condensed" href="/some-url">
        <span class="ffe-button__label">Endre</span>
        <div aria-hidden="true" aria-label="Vennligst vent" class="ffe-button__spinner"></div>
    </a>
    <button class="ffe-button ffe-button--secondary ffe-button--condensed">
        <span class="ffe-button__label">Stopp</span>
        <div aria-hidden="true" aria-label="Vennligst vent" class="ffe-button__spinner"></div>
    </button>
    <button class="ffe-button ffe-button--secondary ffe-button--condensed">
        <span class="ffe-button__label">Slett</span>
        <div aria-hidden="true" aria-label="Vennligst vent" class="ffe-button__spinner"></div>
    </button>
    <a class="ffe-button ffe-button--secondary ffe-button--condensed" href="/some-other-url">
        <span class="ffe-button__label">Flytt til aksjesparekonto</span>
        <div aria-hidden="true" aria-label="Vennligst vent" class="ffe-button__spinner"></div>
    </a>
</div>
```

selbekk:
Definitely a bug - but button text should not wrap!

henrikhermansen:
What's more is `tertiary` buttons doesn't look good in an inline group either.
Just have a look in [the designguide](https://design.sparebank1.no/styleguidist/index.html#!/ButtonGroup/5)

PS! Don't mind the second row, it's actually a `thin` button group, not an `inline` one. My bad!

[...]

PR #266, opened 2018-05-31
fix(ffe-buttons): Stop buttons from breaking

### LINKED ISSUE #264: ffe-buttons: Button text should not wrap

`ffe-buttons`

#### Describe your issue

If a button label becomes long enough it starts wrapping over two lines. The buttons are not designed for this, and the end result is kind of odd looking.

Button labels should never wrap

They
wrap

selbekk:
This commit makes sure our buttons stay on one line. Previously, they
were allowed to break on to two lines - which was not according to design.

Fixes #264

Not sure if this should be a breaking change or not. By the API definition, it's not breaking, but I'm sure there's a few places that would have to have an extra look at their buttons if this is merged.

### DIFF
--- a/packages/ffe-buttons/less/base-button.less
+++ b/packages/ffe-buttons/less/base-button.less
@@ -71,6 +71,7 @@
     text-decoration: none;
     transition: all @ffe-transition-duration @ffe-ease;
     user-select: none;
+    white-space: nowrap;
     width: 100%;
     .ffe-fontsize-button;
 


wkillerud:
IMO if a CSS change has a chance of breaking someones layout it should be a new major version

ivarni (dismissed):

antidecaf (commented):
Can't recall where exactly, but I've seen buttons with several lines of text by design. Maybe a `--text-wrap` modifier would come in handy?

henrikhermansen (dismissed):

henrikhermansen:
This might fix #219 as well, if buttons can no longer wrap their text.

benmatthews (dismissed):

PR #267, opened 2018-05-31
feat(ffe-datepicker-react): Add id prop

selbekk:
This commit passes the id prop to the input field, so that it's easier
to pass the ID to the correct components.

Note: It does NOT auto-generate an ID if one isn't provided. That should be its own discussion imo.

### DIFF
--- a/packages/ffe-datepicker-react/src/datepicker/Datepicker.js
+++ b/packages/ffe-datepicker-react/src/datepicker/Datepicker.js
@@ -284,6 +284,7 @@ export default class Datepicker extends Component {
         const {
             hideErrors,
             inputProps = {},
+            id,
             label,
             language,
             onChange,
@@ -328,6 +329,7 @@ export default class Datepicker extends Component {
                 >
                     <DateInput
                         aria-invalid={this.ariaInvalid()}
+                        id={id}
                         inputProps={inputProps}
                         onBlur={this.onInputBlur}
                         onChange={evt => onChange(evt.target.value)}
@@ -382,10 +384,10 @@ Datepicker.propTypes = {
     ariaInvalid: oneOfType([bool, string]),
     calendarAbove: bool,
     hideErrors: bool,
+    id: string,
     onValidationComplete: func,
-    inputProps: shape({
+    inputProps: shape({ 
         className: string,
-        id: string,
     }),
     label: string,
     language: string,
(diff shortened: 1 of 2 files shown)


wkillerud:
> It does NOT auto-generate an ID if one isn't provided. That should be its own discussion imo.

Apart from bundle size, are there any downsides to generate a random working ID to help developers with a11y? The component already imports `uuid`.

wkillerud (approved):

wkillerud (changes requested):
On second thought - the [`label` element needs an updated `htmlFor`](https://github.com/SpareBank1/designsystem/blob/0e8ebe96e8cf0d8511a2f0c98ca8765657194d4d/packages/ffe-datepicker-react/src/datepicker/Datepicker.js#L312) that also checks for `this.props.id`, otherwise this should have no effect.

ivarni:
But.. it already does this?
[id is already part of inputProps](https://github.com/SpareBank1/designsystem/blob/31c28f6769240e43567535e80096358d7f6f7a25/packages/ffe-datepicker-react/src/datepicker/Datepicker.js#L388) and it gets [destructured and passed to 
 the input](https://github.com/SpareBank1/designsystem/blob/31c28f6769240e43567535e80096358d7f6f7a25/packages/ffe-datepicker-react/src/input/Input.js#L45) already?

I'm all for getting rid of these `inputProps` objects where we can but I'd prefer we don't duplicate them :)

ivarni:
Also, auto-generating an id would not affect the bundle size since [it's already done for the label](https://github.com/SpareBank1/designsystem/blob/31c28f6769240e43567535e80096358d7f6f7a25/packages/ffe-datepicker-react/src/datepicker/Datepicker.js#L27) :)

wkillerud:
> But.. it already does this? id is already part of inputProps and it gets destructured and passed to the input already?

Yes, but the API is kind of clunky. I also mentioned in the issue how it was confusing that if I give the component a `label` prop I had to provide an ID (especially an `inputProps.id`) to get proper linkage between the label and the input field. The PR doesn't address that yet though.

ivarni:
But now we're going to be able to provide an `id` through two different means? I'm not sure how that makes anything less confusing :)

selbekk:
Closing this since abandoned.

PR #268, opened 2018-05-31
chore: Add Prettier as a precommit hook

selbekk:
We always forget about running prettier on our PRs. This commit re-adds
prettier as a pre-commit hook.

It was removed earlier due to some mis-communications I believe.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
         "clean": "run-s clean:lib clean:lerna",
         "clean:lib": "rimraf packages/*/lib",
         "clean:lerna": "lerna clean --yes",
-        "prettier": "pretty-quick --staged",
+        "precommit": "pretty-quick --staged",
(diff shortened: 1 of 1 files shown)


wkillerud:
Maybe add linting and tests on precommit or prepush as well?

wkillerud (approved):

antidecaf (approved):

benmatthews (approved):

selbekk:
I tried adding linting and testing on commit, but it runs all lint tasks and the entire 500+ tests test suite, and that takes ~30 seconds. I say we skip it.

ivarni:
Is 30 seconds per commit really that bad? How often are we commiting stuff in this repo?

selbekk:
Pretty often. 👍 Also, any issues would be caught by CI - so I think we should optimize this one for the developer experience.

PR #269, opened 2018-06-01
149 deprecate hawaii less

ivarni:
Do note that I've not seen a new module being added to this lerna thing before so I'm not sure if this is all the steps that are needed or not.


wkillerud:
> Do note that I've not seen a new module being added to this lerna thing before so I'm not sure if this is all the steps that are needed or not.

Since the component has no documentation relevant for Styleguidist I think this is all we need. Otherwise we'd need a config in [`sections`](https://github.com/SpareBank1/designsystem/blob/431faa9c50f9dd8d1a905ff3a14a6dd0531e65e6/styleguide.config.js#L141).

wkillerud (dismissed):

ivarni:
OK, next question. Does the `ffe-layout` need a *CHANGELOG.md* file or is it generated if it doesn't exist?

wkillerud:
Good question... I see Walters added a CHANGELOG.md with only a title [when he introduced `ffe-webfonts`](https://github.com/SpareBank1/designsystem/commit/f725f32932cbbc2ee2958afc9b3225945215d201#diff-12028a091930ad8a8e689f036903ae7c). I don't know honestly, but I guess it doesn't hurt to add it manually.

selbekk:
What about naming the package ffe-layout-legacy (or similar)?

ivarni:
Sure, I don't care what it's called, I don't plan on using it :)

ivarni:
Next time it would be great if you could give that kind of feedback on the issue (#149) instead, so I don't have to do most of the work twice just for a package-name :)

wkillerud (approved):

selbekk (approved):

PR #272, opened 2018-06-04
feat(ffe-icons): Added trashcan icon to ffe-icons

oleespen:
Added trash-can icon to ffe-icons


packages/ffe-icons/icons/soppelkasse-ikon.svg
```diff
@@ -0,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="200" height="200" viewBox="0 0 200 200">
+  <g fill="none" fill-rule="evenodd" stroke="#005AA4" stroke-width="16.364" transform="translate(10 10)">
+    <path stroke-linecap="round" stroke-linejoin="round" d="M155.454546,65.4545454 L155.454546,163.636364 C155.454546,172.673751 148.128296,180 139.09091,180 L40.909091,180 C31.8717041,180 24.5454546,172.673751 24.5454546,163.636364 L24.5454546,65.4545454 M0,32.7272728 L180,32.7272728"/>
+    <path stroke-linecap="square" d="M90,90 L90,139.09091 M57.2727272,90 L57.2727272,139.09091 M122.727272,90 L122.727272,139.09091 M57.2727272,32.7272728 L57.2727272,16.3636364 C57.2727272,7.32624955 64.5989768,1.66014103e-15 73.6363636,0 L106.363636,0 C115.401023,6.36667091e-15 122.727272,7.32624955 122.727272,16.3636364 L122.727272,32.7272728"/>
+  </g>
+</svg>
```
selbekk: Thanks for adding this!

There's a few things we'd love for you to do before we merge this. First of all, all of our icons that are monochromatic (only have one color) do not have a default fill or stroke color. This makes them more generic and easier to reuse later.

Secondly, we ask that you run SVGs through an optimizer before adding it to the design system. This way, we ensure the least amount of bloat. One such optimizer is Jake Archibald's SVGOMG.

Here's the icon you added, just optimized:

```html
<svg xmlns="http://www.w3.org/2000/svg" width="200" height="200">
  <g fill="none" stroke="#000" stroke-width="16.364">
    <path stroke-linecap="round" stroke-linejoin="round" d="M165.4545 75.4545v98.1819c0 9.0374-7.3262 16.3636-16.3636 16.3636H50.9091c-9.0374 0-16.3636-7.3262-16.3636-16.3636V75.4545M10 42.7273h180"/>
    <path stroke-linecap="square" d="M100 100v49.091M67.2727 100v49.091M132.7273 100v49.091M67.2727 42.7272V26.3636C67.2727 17.3262 74.599 10 83.6364 10h32.7272c9.0374 0 16.3637 7.3262 16.3637 16.3636v16.3637"/>
  </g>
</svg>
````

[Here's a codepen](https://codepen.io/selbekk/pen/zarWXV) you can play with! I noticed that this icon used stroke instead of fill, which is a bit different from the rest of the icons. Not sure how you could move from one to the other, or if it's an issue at all. Right now, I set the stroke value to `#000` to make sure it shows up in the demo.

cc @ivarni - do you know if there's a downside to doing it this way? Or if there's an easy way to move from stroke to fill when it comes to paths?

selbekk (changes requested):

packages/ffe-icons/icons/soppelkasse-ikon.svg
ivarni: What a fascinating way to draw an SVG, I've no idea what's going on there. I'd ask whoever drew it and maybe give them a copy of a few of the existing ones to look at.

Also, please don't include `width` and `height` properties, those should be left to the consumer to control, we just need the `viewBox` here.

oleespen: Hi! Thanks for the feedback, i will check back with our designer who made this :)

selbekk (approved):

PR #273, opened 2018-06-04
fix(ffe-form-react): add expectations and fix tests

henrikhermansen:
Tests must have an expectaion. Simply writing `expect(someBooleanStatement)`
won't actually test anything.

This revealed the tests had another flaw. In order to test the class of an
element returned by a Component returned by the Component we're testing,
we need to mount it to get the full tree.


henrikhermansen:
I tried using `render()` instead of `mount()`. Not sure which is better. I couldn't quite get the results I wanted with either `render()` or `shallow()`, so I believe we must use `mount()` here.

If you have any other suggestions, don't hold back :slightly_smiling_face:

selbekk (approved):
LGTM!

PR #275, opened 2018-06-05
fix: Add button styling to toolbar buttons

antidecaf:
Fixes a bug in which toolbar buttons in the styleguide would be rendered with default browser styles.

Before:
image

After:
image

### DIFF
--- a/src/components/toolbarbutton/ToolbarButtonRenderer.js
+++ b/src/components/toolbarbutton/ToolbarButtonRenderer.js
@@ -4,8 +4,10 @@ import PropTypes from 'prop-types';
 export function ToolbarButtonRenderer({ onClick, href, title, children }) {
     if (href !== undefined) {
         return (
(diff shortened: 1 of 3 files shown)


selbekk:
Would it be okay if we make the icon a tiny bit smaller? Also, a darker shade of blue (like our text)?

selbekk (approved):
Very nice btw!

antidecaf:
Sure it could be smaller, though for some reason that makes the svgs blurry. Example:

20 x 20 px: 
image

24 x 24 px: 
image

The shade of blue is not actually specified in the styleguide css, but a result of adding `ffe-inline-button--tertiary` to the buttons. It's a trivial override though, so not a problem to change it if we want to (?).

selbekk:
Hm that's weird. I say just go ahead and merge - it's an improvement either way.

Issue #254, opened 2018-05-24
There seems to be a problem in the ButtonGroup


ivarni:
#### Which package is this issue related to?

`ffe-buttons` and/or `ffe-buttons-react`

#### Describe your issue (screenshots welcome!)

When looking at the [docs for ButtonGroup](https://design.sparebank1.no/styleguidist/index.html#buttongroup) there seems to be an issue with the inline variant where some of the buttons are vertically challenged when it comes to placement

#### What is the expected behavior? 

![image](https://user-images.githubusercontent.com/379849/40473226-dfe757ca-5f3b-11e8-9877-24816c39aa89.png)

#### What is the actual behavior?

![image](https://user-images.githubusercontent.com/379849/40473201-d18c9ab4-5f3b-11e8-9842-0e9da5b0c06d.png)

ivarni:
Something's up with the small-screen view as well

![image](https://user-images.githubusercontent.com/379849/40473663-1ff591aa-5f3d-11e8-9463-24b0d5c4a408.png)

antidecaf:
Fixed via #276/#277

PR #276, opened 2018-06-05
fix(ffe-buttons): Align buttons in button groups properly

### LINKED ISSUE #219: Buttons in an inline button group doesn't align text correctly

antidecaf:
Fixes #219, as well as a separate issue in which anchors with `.ffe-inline-button--tertiary` styling were not correctly aligned with other buttons in a `.ffe-button-group`.

Before:
image

After:
image

### DIFF
--- a/packages/ffe-buttons/less/button-group.less
+++ b/packages/ffe-buttons/less/button-group.less
@@ -22,6 +22,7 @@
     .ffe-inline-button {
         justify-content: center;
         margin: 0 auto 10px;
+        align-self: center;
     }
 
     &--thin {


benmatthews (approved):

selbekk (approved):

henrikhermansen (approved):

henrikhermansen:
This seems to fix the issue on desktop size in #254 I think?
It won't fix the issue on small size, but I believe @torbein is working on that.

ivarni:
Yeah, I checked, it does indeed seem to fix it for large displays but not small ones, though it does make it better than it is today.

Probably a good idea to make sure the two fixes don't collide.

antidecaf:
Neat. I'll take an extra look to see if that issue can be easily fixed as well while I'm at it.

henrikhermansen:
@antidecaf like I said, @torbein is already looking at that. Have a chat with him to see what he's got :slightly_smiling_face:

antidecaf:
He's on it, I'll merge this one then.

PR #277, opened 2018-06-06
fix(ffe-buttons): Remove left margin on small screen for buttons in b…

torbein:
Removed left margin on small screen for buttons in button-group. 

Before:
button-group-with-left-margin

After:
button-group-without-left-margin


antidecaf (changes requested):
I think a slightly better fix is to simply remove the `&:first-child` selector (line 39). That way there will still be some spacing between inline buttons rendered in a row as opposed to a column.

torbein:
I see. I thought the spacing was supposed to be removed for all buttons. I will make the changes that you suggested.

antidecaf (approved):

Issue #186, opened 2018-04-19
Spinner is shaking in IE11


selbekk:
#### Which package is this issue related to?
ffe-spinner

#### Describe your issue (screenshots welcome!)
The spinner (both small and large) shakes very slightly in IE11. Tried to make a gif of it:

![brrrrrrrr](https://user-images.githubusercontent.com/1307267/38980912-5f8d1148-43be-11e8-97e9-a6c488e97784.gif)

#### What is the expected behavior? 
Not to get epilepsy

#### What is the actual behavior?

![brrrrr](https://media.giphy.com/media/JEGIloZ79M46c/giphy.gif)

#### Steps to reproduce

You can open [this page](https://design.sparebank1.no/styleguidist/index.html#!/Spinner) in IE11 and notice it for yourself

ivarni:
It's been like this for as long as I can remember.

selbekk:
Looks like it's a good ol' IE bug. A quick google search suggested trying to change the transform slightly, to "trick" IE into doing the right thing. ([example](https://stackoverflow.com/questions/42558183/ie-10-11-edge-shaky-on-hover-animation))

benmatthews:
`transform-origin: 50.01%` did not seem to fix the problem... neither does adding some hardware accelerated css like `transform: translate3d(0,0,0)` or ` transform: translateZ(0)`.

I think we could create more fun and interesting spinners by using an inline svg instead of styling a pseudo element.  This would be easy to add to the react packages but might mean a bit more work for people only using the css modules.  I have to admit though I have not tested `transform: rotate` on an svg in IE yet.

selbekk:
@benmatthews unfortunately - IE doesn't support animating SVGs with CSS, so that's kind of a show-stopper :(

ivarni:
What @selbekk said. If I recall correctly we have some SVGs where we've just accepted that animation won't happen in IE but that's not really an option for a spinner.

I'm pretty sure it's possible to do this in IE with just DOM and CSS though, if we wanted to spend the time on it.

wkillerud:
I had a look at this issue today, and from what I could tell moving the animation out from the pseudo-element fixed the shaking for `ffe-spinner`. In other words animating the actual `span` instead of `::before`. Unfortunately it had no effect on the spinner inside the buttons when I did the same thing there (they don't share code) :sob:

PR #283, opened 2018-06-08
ffe-spinner, ffe-buttons - shaky animation in IE

wkillerud:
## Do not merge

Opening a PR to keep track of the progress that's been made in tracking down this issue. The shaking in buttons is not fixed, and the same approach seemingly had no effect there. Also, prettier decided it liked markdown again...

#186

### DIFF
--- a/packages/ffe-buttons/less/base-button.less
+++ b/packages/ffe-buttons/less/base-button.less
@@ -26,7 +26,6 @@
 //
 // Styleguide ffe-buttons.base-button.1
 
-
 // More Buttons
 //
 // Markup:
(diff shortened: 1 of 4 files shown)


packages/ffe-buttons/less/base-button.less
```diff
@@ -168,7 +165,7 @@
     height: 45px;
 }
 
-.ffe-button--expanded {
+.ffe-button--expanded  {
```
ivarni, line 168: :p

ivarni:
@wkillerud du holder fortsatt i denne ikke sant? :D

wkillerud:
Du må gjerne ta over @ivarni 😉

antidecaf:
Er det noen som ser på denne, eller skal vi lukke den?

PR #286, opened 2018-06-13
fix(ffe-icons-react): Don't set inline size on SVG

### LINKED ISSUE #285: React icons are having height/width props set on them

`ffe-icons-react`

When using icons from the `ffe-icons-react` package they all come with the `height` and `width` property set.

I'm expecting us to rely on the FFE components setting only the viewbox, leaving the consumer in charge of controlling the actual size of the icon

During the generation of React components, the size-props are set:
https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-icons-react/src/build-jsx-components.js#L38-L39

In certain cases this forces me to override those sizes using the `style` prop (which will win over these props) but I shouldn't have to be doing that.

ivarni:
No longer adds `width={150} height={150} to the generated SVGs.
This should not affect you, as both inline styling AND CSS will
override those props. The only case where it seems to matter is
if you're embedding an `ffe-icons-react` SVG inside another SVG.

Fixes #285

### DIFF
--- a/packages/ffe-icons-react/src/build-jsx-components.js
+++ b/packages/ffe-icons-react/src/build-jsx-components.js
@@ -8,10 +8,18 @@ mkdirp.sync('./jsx');
 
 const createSvgMap = () => {
     const map = {};
-    const iconsPath = path.join(__dirname, '..', 'node_modules', '@sb1', 'ffe-icons', 'icons');
-    fs.readdirSync(iconsPath)
+    const iconsPath = path.join(
+        __dirname,
+        '..',
+        'node_modules',
+        '@sb1',
+        'ffe-icons',
+        'icons',
+    );
+    fs
+        .readdirSync(iconsPath)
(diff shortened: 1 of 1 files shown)


packages/ffe-icons-react/src/build-jsx-components.js
```diff
@@ -28,21 +36,18 @@ const createSvgMap = () => {
...
-        xmlMode: true
+        xmlMode: true,
     });
     const svg = $('svg');
     // React does not support namespace definitions
     svg.attr('xmlns', null);
     svg.attr('xmlns:svg', null);
 
-    svg.attr('height', 150);
-    svg.attr('width', 150);
```
ivarni, line 39: Looks like the commit-hook went over the formatting, but the only real change is the removal of these lines.

wkillerud (approved):

PR #300, opened 2018-06-18
fix(ffe-core): Fix build issue

### LINKED ISSUE #299: ffe-core doesn't export ES5
ffe-core

When we build ffe-core, it creates some javascript files for variables, which can be imported. Unfortunately, the index file contains a trailing comma for functions, which are not ES5.

This leads to the build process failing whenever you use tools such as `create-react-app`.

selbekk:
This commit fixes an issue where using the JS variables from ffe-core
broke certain builds (in my case, create-react-app). The reason for the
breakage was a trailing comma in a function argument list, which is an
ES6 feature.

This commit removes the trailing comma, and makes the code slightly
easier to read.

Fixes #299

### DIFF
--- a/packages/ffe-core/bin/build.js
+++ b/packages/ffe-core/bin/build.js
@@ -13,21 +13,25 @@ mkdirp.sync(path.resolve(__dirname, '..', 'lib'));
 FILES_WITH_VARIABLES.forEach(filename => {
     const lessContent = fs.readFileSync(
         path.resolve(__dirname, '..', 'less', `${filename}.less`),
-        'utf8'
+        'utf8',
     );
 
     // This is done in a ridiculously step by step fashion to increase
     // readability. Basically, this changes a LESS file into a JavaScript object
-    const variables = lessContent.split('\n')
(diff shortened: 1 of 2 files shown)


ivarni:
I'd much rather see an actual transpile step, or at the very least a big hard-to-miss warning informing the next person of the need to keep the source completely in ES5. Everything else in FFE is written in ES6 so it's not very intuitive that this can't be.

wkillerud:
Ivars suggestion seems more future-proof. This looks OK for now as a quick fix.

wkillerud (dismissed):

selbekk:
Hm.. well, sure, I don't mind adding a transpilation step further down the line.

ivarni:
I don't think this closes the issue but it certainly works as a quickfix. I think leaving the issue open is a good idea so this doesn't re-appear as an issue again later if someone accidentally use new syntax.

selbekk:
I added the babel build step. Good to go?

ivarni (approved):

PR #301, opened 2018-06-18
feat(ffe-core-react): Add `LinkText` component

### LINKED ISSUE #298: ffe-core-react: Implement component for ffe-link-text

`ffe-core-react`

#### Describe your issue

Today we don't have any React component for anchors. The Komponenter list is missing documentation for `.ffe-link-text` because of this.

We should have a component wrapping `.ffe-link-text` for documentation if nothing else.

We don't document `.ffe-link-text`.

wkillerud:
Available as a named export from `ffe-core-react`, `LinkText` is a
wrapper around an `a` element with the proper classes. By default it
renders with an underline, but by setting the `underline` prop to
`false` the component will apply the `--no-underline` modifier.
Otherwise the component supports extending the `className` prop with
custom classes, and `...rest` props are spread directly on the anchor to
support things like `target` and `rel`.

Fixes #298


packages/ffe-core-react/src/typography/LinkText.js
```diff
@@ -0,0 +1,40 @@
...
+    } = props;
+
+    return (
+        <a
+            className={classNames(
+                'ffe-link-text',
+                { 'ffe-link-text--no-underline': !underline },
+                className,
+            )}
+            href={href}
```
ivarni: Only thing I'm curious about is why this couldn't just be passed as part of `...rest`, it doesn't seem to be used for anything.

wkillerud: True, but linting complained about an `a` without an `href` :( A tiny tiny bonus is that we get propType validation during development in case a link from a variable ever becomes null or undefined 😀

packages/ffe-core-react/src/typography/LinkText.js
```diff
@@ -0,0 +1,40 @@
...
+import { bool, node, string } from 'prop-types';
+import classNames from 'classnames';
+
+const LinkText = (props) => {
+    const {
+        children,
+        className,
+        href,
+        underline,
+        ...rest
```
selbekk: I would love for this to have an `element` prop, so that I can use it with `react-router`-s `Link` component.

packages/ffe-core-react/src/typography/LinkText.js
```diff
@@ -0,0 +1,40 @@
...
+        <a
+            className={classNames(
+                'ffe-link-text',
+                { 'ffe-link-text--no-underline': !underline },
+                className,
+            )}
+            href={href}
+            {...rest}
+        >
+            {children}
```
selbekk: You don't have to destructure `children` here - it can be spread with the `rest` prop.

packages/ffe-core-react/src/typography/LinkText.js
selbekk: That also makes the `href` prop unvalidatable - because you don't know whether or not we need a `href` prop or a `to` prop.

selbekk (dismissed):
Great work - especially on the testing and documentation part! 🎉 Would love support for an `element` prop before merging this though

selbekk (commented):
Amazing work bud! ;taco:

selbekk (approved):

PR #302, opened 2018-06-18
feat(ffe-form-react): Add new `description` prop to `InputGroup`

### LINKED ISSUE #291: Add support for always showing a tooltip

`ffe-form-react`

I've been getting several designs lately where the input-group is rendered with an always visible non-toggleable tooltip

image

It would be nice to have support for this out of the box so I can use the already existing `tooltip` prop instead of rendering my own helptext. This would help make sure these stay visiually consistent with the toggleable ones.

For reference, this is what the current ones look like

image

I propose adding a new prop, `ubiquitousTooltip` (no, I'm joking I hate try to spell that word).

Maybe we can call it `staticTooltip` and have it be a `bool`? Setting it would simply remove the ❓ box and always show the tooltip.

ivarni:
Setting `helpText` instead of (or in addition to) the existing `tooltip` renders a static helptext instead of the expandable version. It looks the same, but this one will never leave you.

Fixes #291

### DIFF
--- a/packages/ffe-form-react/src/InputGroup.js
+++ b/packages/ffe-form-react/src/InputGroup.js
@@ -17,6 +17,7 @@ class InputGroup extends Component {
         const {
             children,
             className,
+            description,
             label,
             fieldMessage,
             tooltip,
@@ -39,6 +40,12 @@ class InputGroup extends Component {
             throw new Error(
                 'Only use the "onTooltipToggle" prop if you\'re not sending a component of type ' +
                     '<Tooltip /> in the "tooltip" prop. If you are, use "onClick" on that component instead',
+            );
+        }
+
+        if (tooltip && description) {
+            throw new Error(
+                'Don\'t use both "tooltip" and "description" on an <InputGroup />, pick one of them',
             );
         }
 
@@ -67,8 +74,14 @@ class InputGroup extends Component {
                 {typeof tooltip === 'string' && (
                     <Tooltip onClick={onTooltipToggle}>{tooltip}</Tooltip>
                 )}
+
                 {tooltip === true && <Tooltip onClick={onTooltipToggle} />}
+
                 {React.isValidElement(tooltip) && tooltip}
+
+                {description && (
+                    <div className="ffe-small-text">{description}</div>
+                )}
 
                 {modifiedChildren}
 
@@ -89,6 +102,8 @@ InputGroup.propTypes = {
     className: string,
     /** Use the ErrorFieldMessage component if you need more flexibility in how the content is rendered. */
     fieldMessage: oneOfType([string, node]),
+    /** To just render a static, always visible tooltip, use this. */
+    description: string,
     /** Use the Label component if you need more flexibility in how the content is rendered. */
     label: oneOfType([string, instanceOfComponent(Label)]),
     onTooltipToggle: func,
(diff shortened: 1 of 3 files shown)


packages/ffe-form-react/src/InputGroup.js
```diff
@@ -64,11 +65,15 @@ class InputGroup extends Component {
...
 
-                {typeof tooltip === 'string' && (
-                    <Tooltip onClick={onTooltipToggle}>{tooltip}</Tooltip>
-                )}
-                {tooltip === true && <Tooltip onClick={onTooltipToggle} />}
-                {React.isValidElement(tooltip) && tooltip}
+                {typeof tooltip === 'string' &&
+                    !helpText && (
+                        <Tooltip onClick={onTooltipToggle}>{tooltip}</Tooltip>
+                    )}
```
selbekk: Could we show a warning in the console or throw an error here if the user has provided both the `tooltip` and `helpText` props?  This way, it's easier to spot any potential errors.

selbekk (commented):
The work looks fine. 

I'm a bit wary of adding another prop that's... "almost" equal to an existing one, but doing this in userland is not pretty either:

```js
<InputGroup 
  tooltip={
    <div className="ffe-small-text">Tooltip text goes here</div>
  } 
/>
```

packages/ffe-form-react/src/InputGroup.js
ivarni: To what purpose? Why would anyone ever try to do that? The only reason the code had to care was the even-listener thing.

ivarni:
My initial suggestion was to add something that just made the collapse-functionality from `tooltip` go away but this was suggested instead. The issue (#291) is probably a better place to discuss that.

packages/ffe-form-react/src/InputGroup.js
ivarni: I mean, I could add it, if you really think it's important, but to me it just feels like bloat.

ivarni: Well, to be fair that did make a couple of the checks further down redundant :)

wkillerud (dismissed):

packages/ffe-form-react/src/InputGroup.js
selbekk: It's one of those things that often gets lost in translation, or one prop is left behind when you refactor or whatever. One of the great things I love about React as a library is all their helpful warnings during development. If you wrap the warning code in a dead-code-eliminatable check, there won't be bloat either.

selbekk: Also, some people might try to add both a tooltip AND a help text because of reasons - showing a warning would discourage them from doing so.

ivarni: I've added it and implementet it as the already existing warnings, see later comments.

packages/ffe-form-react/src/InputGroup.md
```diff
@@ -17,13 +17,28 @@ const { Input } = require('.');
         />
     </InputGroup>
 
-    <InputGroup
-        label="E-postadresse"
-        fieldMessage="Ugyldig e-postadresse"
-    >
+    <InputGroup label="E-postadresse" fieldMessage="Ugyldig e-postadresse">
```
ivarni, line 20: screw you prettier 😠

wkillerud (approved):
LGTM

selbekk (approved):
LGTM :)

PR #303, opened 2018-06-19
256 remove colours

ivarni:

### DIFF
--- a/packages/ffe-grid-react/src/GridCol.js
+++ b/packages/ffe-grid-react/src/GridCol.js
@@ -9,7 +9,7 @@ import {
     string,
 } from 'prop-types';
 import classNames from 'classnames';
-import backgroundColors from './background-colors';
+import backgroundColors, { removedColors } from './background-colors';
 
 import {
     checkForDeprecatedModifiers,
@@ -65,10 +65,21 @@ const modifiers = props =>
         .map(key => `ffe-grid__col--${camelCaseToDashCase(key)}`)
         .join(' ');
 
-const backgroundClass = props =>
-    props.background && backgroundColors.includes(props.background)
-        ? `ffe-grid__col--bg-${props.background}`
+const backgroundClass = ({ background }) => {
+    if (!background) {
+        return null;
+    }
+
+    if (removedColors.includes(background)) {
+        throw new Error(
+            `Support for the ${background} background on <GridCol> has been removed, please see the CHANGELOG`,
+        );
+    }
+
+    return backgroundColors.includes(background)
+        ? `ffe-grid__col--bg-${background}`
         : null;
+};
 
 export default class GridCol extends Component {
     componentDidMount() {
@@ -126,13 +137,10 @@ GridCol.defaultProps = {
 GridCol.propTypes = {
     /** Supported background colors */
     background: oneOf([
-        'blue-cobalt',
         'blue-ice',
         'blue-pale',
-        'blue-royal',
         'green-mint',
         'grey-cloud',
-        'purple-magenta',
         'sand',
         'grey-warm',
         'orange-salmon',
--- a/packages/ffe-grid-react/src/GridRow.js
+++ b/packages/ffe-grid-react/src/GridRow.js
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import { bool, node, oneOf, string } from 'prop-types';
 import classNames from 'classnames';
 
-import backgroundColors from './background-colors';
+import backgroundColors, { removedColors } from './background-colors';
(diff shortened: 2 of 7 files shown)


wkillerud (dismissed):
🎉

ivarni:
@wkillerud @selbekk @antidecaf Can we please relax the coveralls-check slightly? A *0.07%* decrease in testcoverage is possibly simply from a rounding error or something and shouldn't prevent a merge in my opinion. Can't it be thresholded at 1% or something?

wkillerud:
It likely complains about missing coverage for the _error throwing if using a removed color_ case.

I'm wary of relaxing the requirement since it opens us to "death by a thousand cuts". That's just my two cents though, if there's consensus for relaxing the coverage check then by all means :)

ivarni:
I kinda figured we didn't need a test for that check since I just added it to pre-empt any comments on it like I got on my other PR. I suppose I could add a test if that's what it takes to get pass the coveralls thing...

packages/ffe-grid-react/src/GridCol.js
```diff
@@ -9,7 +9,7 @@ import {
     string,
 } from 'prop-types';
 import classNames from 'classnames';
-import backgroundColors from './background-colors';
+import backgroundColors, { removedColours } from './background-colors';
```
selbekk: I think we use the American spelling of color in FFE. Keeping this consistent will also improve my own sanity. ;-*

packages/ffe-grid-react/src/GridCol.js
```diff
@@ -65,10 +65,21 @@ const modifiers = props =>
...
+        return null;
+    }
+
+    if (removedColours.includes(background)) {
+        throw new Error(
+            `Support for the ${background} background has been removed, please see the CHANGELOG`,
+        );
+    }
+
+    return backgroundColors.includes(background)
```
selbekk: Perhaps you can throw an error explaining to the user that an invalid color was provided here as well? This way, it's easier to spot the error!

packages/ffe-grid-react/src/GridRow.js
```diff
@@ -27,11 +27,18 @@ export default class GridRow extends Component {
...
         const hasBackgroundColor = backgroundColors.includes(background);
+        const hasRemovedColour = removedColours.includes(background);
 
         if (hasBackgroundColor) {
             content = <div className="ffe-grid__row-wrapper">{children}</div>;
         }
 
+        if (hasRemovedColour) {
+            throw new Error(
+                `Support for the ${background} background has been removed, please see the CHANGELOG`,
```
selbekk: Please specify that it's the ffe-grid package you'll have to investigate - it might not be obvious to all users depending on context.

selbekk (changes requested):
Great job! 🎉 Would be great if you could fix the color-spelling (which - apart from obviously being the right way of spelling that word - also is the standard of our design system), otherwise 👍

packages/ffe-grid-react/src/GridCol.js
ivarni: I've already added that..

packages/ffe-grid-react/src/GridRow.js
ivarni: I'll add the name of the component

packages/ffe-grid-react/src/GridCol.js
selbekk: You've added it for removed colors - but if I passed in the string `blue-fale` instead of `blue-pale`, I'd just get a white background without any indication where the error was.

packages/ffe-grid-react/src/GridCol.js
ivarni: Actually no, we use `grey` over `gray` already but if that's what it takes to get this thing off my TODO list I'll change it.

selbekk: Well, I agree we should be consistent on that level as well - but we should _at least_ use the same spelling of the same word :P

packages/ffe-grid-react/src/GridCol.js
ivarni: The propTypes check already warn of that, having two errors thrown seem utterly pointless. I can see the case for doing both for a colour that was suddenly removed from the supported list but I simply can't see the point of doing it for every unsupported string of text.

selbekk (approved):

selbekk:
Good stuff buddy!

packages/ffe-grid-react/src/GridCol.js
selbekk: Ah didn't think of the prop types. Good point.

PR #305, opened 2018-06-20
fix(ffe-accordion): Added peerDependency to ffe-core@^13.0.0

benmatthews:
A breaking change was made to ffe-core@13.0.0 as the Hawaii layout system was moved to its own
package.  None of these modules are using the Hawaii layout system so should be compatible with the
latest version of ffe-core.

### DIFF
--- a/packages/ffe-accordion/package.json
+++ b/packages/ffe-accordion/package.json
@@ -22,7 +22,7 @@
     "@sb1/ffe-core": "^13.0.1"
   },
   "peerDependencies": {
-    "@sb1/ffe-core": "^12.0.0"
+    "@sb1/ffe-core": "^12.0.0 || ^13.0.0"
   },
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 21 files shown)


benmatthews:
Hmm should I have made 1 commit per package?

packages/ffe-accordion/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sb1/ffe-accordion",
-  "version": "4.0.12",
+  "version": "4.0.13",
```
wkillerud: `lerna` will update this version number automatically, so please don't modify it here

wkillerud (changes requested):
I doubt `lerna` will pick up all the scopes and titles from that one single commit. It'll be more work, but please add the packages in separate commits - one per package (copy-paste is your friend here for the title and body).

ivarni:
Actually, scripting is your friend, not copypaste :)

wkillerud (approved):
Awesome, thanks! 🚀

Issue #306, opened 2018-06-20
peerDependency warnings after upgrading to ffe-icons 6.x


benmatthews:
#### Which package is this issue related to?
All packages with a dependency on `ffe-icons-react@^5.0.0` specifically `ffe-system-messages-react` 

#### Describe your issue (screenshots welcome!)
The latest version of `ffe-system-messages-react` depends on `ffe-icons-react@^5.0.0` but there is now a new major version.

#### What is the expected behavior? 
I want to be able to use the latest version of `ffe-icons-react` in my application as well as `ffe-system-message-react`

#### What is the actual behavior?
Peer dependency warnings as my application is using `ffe-icons-react@^6.0.0

#### Steps to reproduce
Use the latest version of `ffe-system-message-react` and `ffe-icons-react` and checkout the peer dependency warnings in the console.

benmatthews:
fixed in #307

PR #307, opened 2018-06-20
fix(ffe-system-message-react): Updated to ffe-icons-react@^6.0.0

benmatthews:
fix(ffe-system-message-react): added peerDependency to ffe-icons-react@^6.0.0

### DIFF
--- a/packages/ffe-system-message-react/package.json
+++ b/packages/ffe-system-message-react/package.json
@@ -36,7 +36,7 @@
     "react-dom": "^16.2.0"
   },
   "peerDependencies": {
-    "@sb1/ffe-icons-react": "^5.0.0",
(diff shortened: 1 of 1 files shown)


benmatthews:
link to issue #306

wkillerud:
Hmm, I'm not sure this needs to be a breaking change. We know the component works well with 5.x of `ffe-icons-react`, and if it looks OK with `6.x` as well then both majors would be valid peerDependencies.

In other words I think we can say `5.x || 6.x` and make it a patch.

benmatthews:
@wkillerud I was kinda thinking the same thing myself, but then you could have applied that logic to the versioning of `ffe-icons-react` also.  Anyway I'm all for just bumping the patch version as it makes life simpler... will update the PR.

wkillerud (approved):

PR #337, opened 2018-06-29
feat: Change tab styling to resemble actual tabs, rename components

antidecaf:
BREAKING CHANGE: This commit renames the preexisting `Tab` and `TabGroup`components to `TabButton` and `TabButtonGroup`. A new `Tab` and `TabGroup` component is added, but uses different styling in order to represent actual tabs.

The corresponding css classes are renamed from `.ffe-tab` and `.ffe-tab-group` to `.ffe-tab-button` and `.ffe-tab-button-group`, while the old classnames are used for the new look.

The new components does not have the `ghost` prop for transparent styling.

The new components render tabs looking like this:

image

As opposed to the old/renamed version:

image

To upgrade to this release, rename all instances of `<Tab>` and `<TabGroup>` to `<TabButton>` and `<TabButtonGroup>` if you wish to keep using the old version.

If you want to switch to the new look, the markup remains the same as in previous versions. However, the new `Tab` component must be used in the context of a `TabGroup`.


wkillerud:
> If you want to switch to the new look, no changes need to be made unless you use the ghost prop, in which case it should simply be removed.

This is probably not true since the surrounding layout likely won't work without come changes. I think we should skip that sentence.

antidecaf:
@wkillerud True. I'll try to add something about having to consider the surrounding layout.

packages/ffe-tabs-react/src/TabButton.md
```diff
@@ -0,0 +1,44 @@
+# Tab buttons
```
wkillerud: Styleguidist will render a header for us, so this should be removed (otherwise it will render as a duplicate h1)

packages/ffe-tabs-react/src/Tab.md
```diff
@@ -3,50 +3,40 @@
 Tabs er en komponent for å lage knapper med fane-funksjonalitet - gruppering av innhold der én gruppe er synlig og de øvrige er skjult.
 
 ```js
-<Tab>
-    Dette er en tab button
-</Tab>
+<Tab>Dette er en tab</Tab>
```
wkillerud: Mulig vi vil beskrive det som en _fane_ i dokumentasjonen, ikke en _tab_?

wkillerud (dismissed):
Nice! Some minor comments, but this looks great!

packages/ffe-tabs-react/src/Tab.md
antidecaf: Ja, enig.

wkillerud:
Is Github confusing me or does the commit message not include the BREAKING CHANGE message? It needs to be in the commit message body or else Lerna will just release it as a minor version.

antidecaf:
@wkillerud You're right, I added it to the message body of the PR. I'll fix the commit message.

packages/ffe-tabs-react/src/Tab.md
```diff
@@ -1,52 +1,40 @@
...
-                </Tab>
+                <Tab condensed={true}>Dette er en kompakt fane</Tab>
             </div>
         </GridCol>
     </GridRow>
-</Grid>
+</Grid>;
 ```
 
 I tillegg kan man sende inn egendefinerte props for å legge til clickhandlere andre attributter:
```
hdaljord: På flisespikkerisiden så mangler det et 'og' i denne setningen :)

hdaljord (commented):
Looks very good to me!

packages/ffe-tabs-react/src/Tab.md
hdaljord: Dårlig gjort å gi deg skylden for en setning som sto der fra før. Setter den til approved :)

hdaljord (dismissed):

packages/ffe-tabs-react/src/Tab.md
antidecaf: Godt observert! Jeg fikser det. :)

adidick-zz (commented):
Sounds like a good idea to me!

antidecaf:
Build is currently breaking, due to [eslint-plugin-jsx-a11y/issues/454](https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/454) as far as I can tell. [A fix is on the way](https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/456).

PR #348, opened 2018-07-12
fix(ffe-grid): Give all grid-col descendants max-width 100%

ivarni:
BREAKING CHANGE: All direct descendants of a grid column now
receives `max-width: 100%`. This fixes an issue in IE where
the contents of centered grid columns would ignore the width
of the column. This might break parts of your layout, e.g. if
you have SVGs directly inside full-width centered columns (they
will now use the entire width where they might not have done so
before) and possibly also in other scenarios


selbekk (approved):
LGTM

PR #352, opened 2018-07-13
fix(ffe-buttons(-react)): Fixes styling for safari 10

### LINKED ISSUE #350: Incorrect button design on safari 10
ffe-buttons (maybe ffe-buttons-react)

- Button label text is not centered
- Tertiary's `::after` border is positioned next to the button label

Centering may be related to: https://github.com/philipwalton/flexbugs/issues/236

Design is consistent across all supported devices
Design is not consistent on Safari 10
#### Steps to reproduce
Go to http://design.sparebank1.no/styleguidist/index.html#bruk on a Safari 10 browser

AndersCan:
Fixes https://github.com/SpareBank1/designsystem/issues/350
I'm no CSS expert, so please come with all comments and suggestions.

tertiary border: Solved by setting border on label, not `::after` element

centering: fixed by adding a wrapper container to the label: `<div className="ffe-button__label-container"> ...</div>`

### DIFF
--- a/packages/ffe-buttons-react/src/BaseButton.js
+++ b/packages/ffe-buttons-react/src/BaseButton.js
@@ -39,17 +39,21 @@ const BaseButton = props => {
             ref={innerRef}
             {...rest}
         >
-            <span className="ffe-button__label">
-                {leftIcon &&
-                    React.cloneElement(leftIcon, {
-                        className: 'ffe-button__icon ffe-button__icon--left',
-                    })}
-                {children}
-                {rightIcon &&
-                    React.cloneElement(rightIcon, {
-                        className: 'ffe-button__icon ffe-button__icon--right',
-                    })}
-            </span>
+            <div className="ffe-button__label-container">
+                <span className="ffe-button__label">
+                    {leftIcon &&
+                        React.cloneElement(leftIcon, {
+                            className:
+                                'ffe-button__icon ffe-button__icon--left',
+                        })}
+                    {children}
+                    {rightIcon &&
+                        React.cloneElement(rightIcon, {
+                            className:
+                                'ffe-button__icon ffe-button__icon--right',
+                        })}
+                </span>
+            </div>
             {supportsSpinner && (
                 <div
                     aria-hidden={!isLoading}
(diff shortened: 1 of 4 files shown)


packages/ffe-buttons-react/src/BaseButton.js
```diff
@@ -39,17 +39,21 @@ const BaseButton = props => {
...
-                    React.cloneElement(leftIcon, {
-                        className: 'ffe-button__icon ffe-button__icon--left',
-                    })}
-                {children}
-                {rightIcon &&
-                    React.cloneElement(rightIcon, {
-                        className: 'ffe-button__icon ffe-button__icon--right',
-                    })}
-            </span>
+            <div className="ffe-button__label-container">
```
selbekk, line 42: I am inherently sceptical of this move of adding a container-div inside of the actual button, in order to fix a fixed, non-critical bug in a deprecated version of iOS. 

I mean, I'm sure it fixes the issue, but adding another element to the DOM like this, in one of our most widely reused components, can be prone to regressions. Also - the fewer divs we render, the better it is IMO.

packages/ffe-buttons/less/inline-base-button.less
```diff
@@ -71,10 +71,8 @@
     color: @ffe-blue-azure;
     flex-wrap: wrap;
 
-    &::after {
+    .ffe-inline-button__label {
```
selbekk, line 74: There was some reason I decided to solve this by means of a pseudo element - but I can't for the life of me remember what it was.

selbekk (commented):
I'd like if somebody else took a look before we merged this as well.

packages/ffe-buttons-react/src/BaseButton.js
AndersCan: This fix isn't critical for me either and I can remove it.

http://gs.statcounter.com/ios-version-market-share/mobile/norway/#monthly-201706-201806 says  >7% anyway.

AndersCan:
As this is a bug with older iOS version with >7% usage, I've decided to just to close the issue.

ivarni:
Here's me hoping you meant **<** 7% usage :p

PR #354, opened 2018-07-17
fix(ffe-core): Lock line-height for ffe-link-text

### LINKED ISSUE #308: Should we set a line-height on anchors?

`ffe-core` (typography)

When rendering a `.ffe-link-text` element directly inside a flex-container we get a lot of whitespace between the underline (implementet as a border) and the text:

This is because while flex-box can be incredible helpful it sometimes decides to turn into an angry vicious demonspawn.

Setting `line-height: 1em;` will fix this, so I'm thinking we can fix this by just setting that as a part of the `.ffe-link-text` class? It should behave as expected for all font-sizes because *em*s.

I expect the underline on our links to always have the same distance from the text it underlines.

image

image

#### Steps to reproduce

Render an element with `.ffe-link-text` as a direct child of a flexcontainer. The underline will be too far below the text. Now either wrap it in a `<div>` so it's not the direct child **or** give it a `line-height: 1em` CSS rule and it works as expected

ivarni:
Setting the `line-height` property to `1em` for `ffe-link-text` ensures
the correct distance between the text and the underline, also when the
link element is rendered directly in a flex-container.

Fixes #308

### DIFF
--- a/packages/ffe-core/less/typography.less
+++ b/packages/ffe-core/less/typography.less
@@ -41,7 +41,6 @@
 //
 // Styleguide ffe-core.typography.headings-modifiers
 
-
 // Body paragraph
 //
 // Markup:
@@ -279,6 +278,7 @@
     cursor: pointer;
     text-decoration: none;
     word-wrap: break-word;
+    line-height: 1em;
 
     &:hover {
         border-bottom-color: @ffe-blue-cobalt;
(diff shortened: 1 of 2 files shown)


selbekk (approved):
Very clever!

antidecaf (approved):

PR #362, opened 2018-08-09
Remove inline JavaScript from less

kwltrs:
This change enables to use less 3 without having to allow inline JavaScript (`lessc --js` and `javascriptEnabled` option).


antidecaf (dismissed):

antidecaf (dismissed):

kwltrs:
Changed commit message of 5066cedf5ebd7ea38122a9b4a090ae9d176f643e to make it an fix instead of feature.

PR #365, opened 2018-08-13
Update eslint peerdep

kwltrs:
Greenkeeper didn't upgrade the peer dependency of the ffe-linting packages to eslint.


ivarni (dismissed):

selbekk (dismissed):
Should we kill the eslint backward support soon?

antidecaf (dismissed):

kwltrs:
I changed the OR-ing of the peerDependencies entries to a greater-than-range. Currently our linting packages support these older versions of eslint and plugins, so I don't see any need to drop backward support yet.

antidecaf (approved):

PR #369, opened 2018-08-13
feat(ffe-account-selector-react): Migrate from deprecated componentWi…

eivinwi:
Migrated from the now deprecated componentWillReceiveProps lifecycle hook to getDerivedStateFromProps for the AccountSelector. No functionality should be affected by this change.

### DIFF
--- a/packages/ffe-account-selector-react/src/subcomponents/input-field/InputField.js
+++ b/packages/ffe-account-selector-react/src/subcomponents/input-field/InputField.js
@@ -19,6 +19,13 @@ class Input extends Component {
         };
     }
 
+    static getDerivedStateFromProps(nextProps, prevState) {
+        if (!prevState.isFocused || prevState.value !== nextProps.value) {
+            return { value: nextProps.value };
+        }
+        return null;
+    }
+
     onChange(e) {
         const value = e.target.value;
         this.setState({ value });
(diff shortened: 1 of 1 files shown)


selbekk (dismissed):

selbekk:
Note that this will break for any users of React < 16.3. That should either be a breaking change, or we should start using this: https://github.com/reactjs/react-lifecycles-compat

What do you think?

kwltrs:
An upgrade to react 17 should be an all-in-big-bang operation. So let's wait with this one.

kwltrs:
@eivinwi Could you explain why this change is necessary?

selbekk:
A possible reason for wanting this is removing warnings from the [`<StrictMode />`](https://reactjs.org/docs/strict-mode.html) component, which is something you can use to make sure you're mega-compliant with any new changes.

TBH, I don't really see the downside of wrapping our components in the `react-lifecycle-compat` HOC if that helps us migrate step by step. We'll probably need to add it at some point anyhow.

eivinwi:
This should have been a breaking change, you are correct!  

Should I pull the PR, or amend it with the compat-polyfill? The change is mostly about future-proofing the code, so waiting until this is changed accross the board is fine by me.

selbekk:
I would suggest you look into introducing the compat polyfill. I've created [an issue](https://github.com/SpareBank1/designsystem/issues/372) to do the same for all other packages.

kwltrs:
I see the point in having the changes of this pull request. But I prefer to wait with it until we have agreed on an upgrade path to react 17.

selbekk:
I'm not sure what you mean by an upgrade path - perhaps you can create an issue explaining your thoughts, and your suggested path of action?

kwltrs:
@selbekk Let's move the discussion to #372

ivarni:
Is anyone working on this or can it be closed?

ivarni:
I'll close this since it seem inactive, if someone picks up work on it again, we can reopen it maybe?

PR #407, opened 2018-09-18
Refactor focus outline

### LINKED ISSUE #11: Contrast on focus ring is low for certain elements, backgrounds
See attached images. Can we change what color we use for the focus ring?

color-contrast-blue-background
color-contrast-radio-switch

antidecaf:
This PR adjusts the focus outline on a number of form elements, according to the mockups provided in #11.

#### Screenshot of buttons in focus on different backgrounds (couldn't be bothered to switch text colors for all examples, but you get the gist of it):

focus-buttons

#### Screenshot of other form elements with the updated focus outline:

focus-form

Fixes #11

### DIFF
--- a/packages/ffe-core/less/colors.less
+++ b/packages/ffe-core/less/colors.less
@@ -1,47 +1,47 @@
 // Blue
 @ffe-blue-royal: #002776; // Titles, lead, overlays
-@ffe-blue-cobalt: #005AA4; // Primary buttons, emphasis on blue areas and product cards
-@ffe-blue-azure: #0071CD; // Secondary buttons
-@ffe-blue-sky: #7FC6E8;
-@ffe-blue-pale: #DFF1F9; // Info-boxes, etc.
-@ffe-blue-ice: #EFF8FC; // Expandable table rows
-@ffe-blue-focus: rgba(68, 192, 255, 0.5); // Focus on buttons and controls
+@ffe-blue-cobalt: #005aa4; // Primary buttons, emphasis on blue areas and product cards
+@ffe-blue-azure: #0071cd; // Secondary buttons
+@ffe-blue-sky: #7fc6e8;
+@ffe-blue-pale: #dff1f9; // Info-boxes, etc.
+@ffe-blue-ice: #eff8fc; // Expandable table rows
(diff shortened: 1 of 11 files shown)


kwltrs (dismissed):

hodoea:
Based on Adis mockups, the color of the focus outline is lighter..See screenshot. skjermbilde 2018-09-18 kl 16 01 18

Yours is darker:
skjermbilde 2018-09-18 kl 16 01 09

antidecaf:
@hodoea You're right, there's an alpha transparency in `@ffe-blue-focus` for some reason, making it appear darker on darker backgrounds. If I remove the transparency the focus outline appears a bit darker than it should, as far as I can tell. Should we update the color while we're at it? @adidick What color is the outline in your mockups?

Here's how it would look without the transparency:

image image

adidick-zz:
Looks good and seems to work well on all types of backgrounds. 

@antidecaf 
The focus colour I've used in the mockups is: #A1DFFF
That probably means we should update @ffe-blue-focus to that as well (instead of #44C0FF as it is now).

antidecaf:
@adidick Looks good, I've updated the color. Here's `#A1DFFF`, for reference:

image image

packages/ffe-core/less/colors.less
```diff
@@ -1,47 +1,47 @@
...
-@ffe-blue-sky: #7FC6E8;
-@ffe-blue-pale: #DFF1F9; // Info-boxes, etc.
-@ffe-blue-ice: #EFF8FC; // Expandable table rows
-@ffe-blue-focus: rgba(68, 192, 255, 0.5); // Focus on buttons and controls
+@ffe-blue-cobalt: #005aa4; // Primary buttons, emphasis on blue areas and product cards
+@ffe-blue-azure: #0071cd; // Secondary buttons
+@ffe-blue-sky: #7fc6e8;
+@ffe-blue-pale: #dff1f9; // Info-boxes, etc.
+@ffe-blue-ice: #eff8fc; // Expandable table rows
+@ffe-blue-focus: #a1dfff; // Focus on buttons and controls
```
antidecaf, line 8: Prettier... :confused: This line is the only actual change in this file.

adidick-zz (approved):
Pretty!

PR #412, opened 2018-09-19
Added missing border on invalid searchable dropdown

pethel:
I was missing this read border when aria-invalid="true". Most of the other components are wrapped in an <InputGroup /> but as this component provide its own label I suppose it should not be used together with <InputGroup />. Documentation did not show usage with <InputGroup /> either

### DIFF
--- a/packages/ffe-searchable-dropdown-react/less/searchable-dropdown.less
+++ b/packages/ffe-searchable-dropdown-react/less/searchable-dropdown.less
@@ -74,4 +74,8 @@
       border-top: none;
     }
   }
+  [aria-invalid='true'] {
+    border-color: @ffe-orange-fire;
+    border-style: solid;
+  }
 }


antidecaf:
Thanks for your contribution! The addition looks good on its own, however the pattern in other form elements, including regular dropdowns, is a bit different. Styling on `aria-invalid` depends on the `.ffe-input-group` container, which seems to be missing in the searchable dropdown component. I think I would prefer to modify the `ffe-searchable-dropdown-react` API to add the missing classname to the container, rather than add the same styling to a slightly different markup structure.

Thoughts?

ivarni:
In my opinion the component should not bring it's own label but should rely on being wrapped in a `<InputGroup>` component for that. It's one of the reasons why I'm not using this package.

But seeing as it already *does* roll its own everything, I don't see a problem with fixing this issue. Making it more of a simple component is a separate concern in my opinion. The API is what it is, and should work as expected unless it's changed to just render the actual dropdown.

pethel:
+1 on @ivarni

antidecaf:
Yeah, I kind of agree. Upon further inspection the pattern might not be as clear cut either.

Another thing: The commit message needs to [include scope/package name](https://github.com/SpareBank1/designsystem/blob/develop/CONTRIBUTING.md#commit-meldinger) for the fix to be published as a new version. It should be rewritten to something like this:
`fix(ffe-searchable-dropdown-react): Added missing border on invalid searchable dropdown`

pethel:
@antidecaf Thanks I changed the commit message

Issue #416, opened 2018-09-21
Wrong usage of setState


pethel:
This is from the documentation on InlineExpandButton. 

```
import { InlineExpandButton } from '@sb1/ffe-buttons-react';

onClick={() => setState({ isExpanded: !state.isExpanded })}

```

setState is asynchronous. 

I think the correct usage should be 

```
import { InlineExpandButton } from '@sb1/ffe-buttons-react';

onClick={() => setState(prevState => ({ isExpanded: !prevState.isExpanded }))}

```

henrikhermansen:
Indeed, this is the recommended way to update state based on previous state :+1:

PR #417, opened 2018-09-21
Setstate used incorrectly

pethel:
Denna pull request løser føljande issue som jag rapporterte
https://github.com/SpareBank1/designsystem/issues/416

### DIFF
--- a/packages/ffe-accordion-react/src/AccordionItem.js
+++ b/packages/ffe-accordion-react/src/AccordionItem.js
@@ -36,7 +36,7 @@ class AccordionItem extends Component {
     toggle = () => {
         const open = this.getOpen();
         this.props.onToggleOpen(!open);
-        this.setState({ open: !open });
+        this.setState(prevState => ({ open: !prevState.open }));
     };
 
     render() {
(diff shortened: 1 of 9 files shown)


packages/ffe-accordion-react/src/AccordionItem.js
```diff
@@ -36,7 +36,7 @@ class AccordionItem extends Component {
     toggle = () => {
         const open = this.getOpen();
         this.props.onToggleOpen(!open);
-        this.setState({ open: !open });
+        this.setState(prevState => ({ open: !prevState.open }));
```
benmatthews, line 39: Is there a chance that `open` returned by `this.getOpen()` can get out of sync with `prevState.open` ?

For example if `toggle` is called many times when `setState`will queue up the changes?

benmatthews (approved):

packages/ffe-accordion-react/src/AccordionItem.js
pethel: Good catch. Should be fixed. We should not use other state then prevState(and currentProps) to calculate the next state

kwltrs (approved):

PR #422, opened 2018-09-24
fix(ffe-buttons): Specify buttongroup width in order to support full width

### LINKED ISSUE #418: ButtonGroup does not use fill widht
`<ButtonGroup />` does not use full width on mobile when placed in a standard `<GridCol/>`. In this case its because the `<GridCol/>` has a `display: flex`. In the documentation this is not an issue. 

Is this something that should be fixed? I would assume it standard to use the `<ButtonGroup />` as child of `<GridCol/>` in real life applications. 

image

antidecaf:
This a fix to an issue in which buttons wrapped in `.ffe-button-group` would not be rendered in full width on mobile screens. This behaviour differs from `.ffe-button`, which is displayed as 100% wide by default on small screens.

#### Before:
image

#### After:
image

Fixes #418

### DIFF
--- a/packages/ffe-buttons/less/button-group.less
+++ b/packages/ffe-buttons/less/button-group.less
@@ -17,6 +17,7 @@
     display: flex;
     flex-direction: column;
     padding: 40px 0;
+    width: 100%;
 
     .ffe-button,
     .ffe-inline-button {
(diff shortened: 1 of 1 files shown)


selbekk (commented):
Looks good!

selbekk (approved):

PR #423, opened 2018-09-25
feat(ffe-searchable-dropdown-react): Extended noMatch to allow for a …

patrisk:
…set of fallback items.

This commit should be backwards compatible while allowing for a new set of items to be displayed for when the search results fail.

The appearance of the dropdown before there is an interaction should remain the same, while the dropdown with this extended noMatch feature will appear like this:
skjermbilde

### DIFF
--- a/packages/ffe-searchable-dropdown-react/src/ListContainer.js
+++ b/packages/ffe-searchable-dropdown-react/src/ListContainer.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { array, string, number, func } from 'prop-types';
+import { array, string, number, func, oneOfType, shape } from 'prop-types';
 import ListItem from './ListItem';
 
 const ListContainer = ({
@@ -13,9 +13,13 @@ const ListContainer = ({
     maxRenderedDropdownElements,
 }) => {
     const itemsToRender =
-        maxRenderedDropdownElements > 0
-            ? dropdownList.slice(0, maxRenderedDropdownElements)
-            : dropdownList;
+        dropdownList.length > 0
+            ? maxRenderedDropdownElements > 0
+                ? dropdownList.slice(0, maxRenderedDropdownElements)
+                : dropdownList
+            : typeof noMatch === 'string'
+                ? []
+                : noMatch.items;
     return (
         <ul
             className="ffe-searchable-dropdown__scroll-container-list"
@@ -26,6 +30,12 @@ const ListContainer = ({
                 highlightedIndex
             }
         >
+            {dropdownList.length === 0 && (
+                <li className="ffe-searchable-dropdown__item">
+                    {typeof noMatch === 'string' ? noMatch : noMatch.text}
+                </li>
+            )}
+
             {itemsToRender.map((item, index) => {
                 return (
                     <ListItem
@@ -44,9 +54,6 @@ const ListContainer = ({
                 dropdownList.length > maxRenderedDropdownElements && (
                     <li className="ffe-searchable-dropdown__item">...</li>
                 )}
-            {dropdownList.length === 0 && (
-                <li className="ffe-searchable-dropdown__item">{noMatch}</li>
-            )}
         </ul>
     );
 };
@@ -56,7 +63,13 @@ ListContainer.propTypes = {
     dropdownList: array.isRequired,
     maxRenderedDropdownElements: number,
     highlightedIndex: number.isRequired,
-    noMatch: string.isRequired,
+    noMatch: oneOfType([
+        string,
+        shape({
+            text: string.isRequired,
+            items: array,
+        }),
+    ]).isRequired,
     onSelect: func.isRequired,
     renderDropdownElement: func,
     refHighlightedListItem: func.isRequired,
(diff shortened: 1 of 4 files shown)


packages/ffe-searchable-dropdown-react/src/ListContainer.js
```diff
@@ -13,9 +13,13 @@ const ListContainer = ({
     maxRenderedDropdownElements,
 }) => {
     const itemsToRender =
-        maxRenderedDropdownElements > 0
-            ? dropdownList.slice(0, maxRenderedDropdownElements)
-            : dropdownList;
+        dropdownList.length > 0
```
pethel, line 16: You have structured the lines in a nice way, but personly I am not a fan of nesting tenary.

packages/ffe-searchable-dropdown-react/src/SearchableDropdown.js
```diff
@@ -264,8 +274,15 @@ SearchableDropdown.propTypes = {
...
     inputValue: string,
-    /** value to be displayed in dropdown in case of search with no match */
-    noMatch: string.isRequired,
+    noMatch: oneOfType([
+        /** value to be displayed in dropdown in case of search with no match */
+        string.isRequired,
+        /** or value to be display along another list of items */
+        shape({
+            text: string.isRequired,
+            items: [],
```
pethel: ` items: array, ` eller ` items: arrayOf`?

packages/ffe-searchable-dropdown-react/src/ListContainer.spec.js
```diff
@@ -127,6 +127,22 @@ describe('<ListContainer>', () => {
...
+            ...defaultProps,
+            noMatch: {
+                text: noMatchText,
+                items: yrker.slice(0, 4),
+            },
+            dropdownList: [],
+        };
+        const component = mountDropdownWithProps(props);
+        expect(component.find(ListItem)).toHaveLength(4);
+        expect(
```
pethel: I think this should be possible

```
import ListItem from './ListItem.js' 
 component.find(ListItem).text(),
```
I my opinio it makes the test less fragile

packages/ffe-searchable-dropdown-react/src/ListContainer.js
```diff
@@ -56,7 +63,13 @@ ListContainer.propTypes = {
     dropdownList: array.isRequired,
     maxRenderedDropdownElements: number,
     highlightedIndex: number.isRequired,
-    noMatch: string.isRequired,
+    noMatch: oneOfType([
+        string.isRequired,
+        shape({
+            text: string.isRequired,
+            items: [],
```
pethel: I am not aware of using an array like that for prop-types
https://www.npmjs.com/package/prop-types

packages/ffe-searchable-dropdown-react/src/SearchableDropdown.js
patrisk: You're right, this one slipped my mind. I suppose we would want array, since we can't necessarily know the structure of the data? 

Edit: I suppose we could use arrayOf(object). The intended goal is to support the same item structure as the rest of the drop down items, so I think using arrayOf(object) in SearchableDropDown and array in ListContainer would be the better choice to minimize any confusion?

packages/ffe-searchable-dropdown-react/src/ListContainer.js
patrisk: No, this was a mistake on my part.

packages/ffe-searchable-dropdown-react/src/ListContainer.js
patrisk: What do you think about something like this:

```js
    let itemsToRender = maxRenderedDropdownElements > 0
            ? dropdownList.slice(0, maxRenderedDropdownElements)
            : dropdownList;

    if (dropdownList.length === 0) {
        itemsToRender = typeof noMatch === 'string'
            ? []
            : noMatch.items;
    }
```

packages/ffe-searchable-dropdown-react/src/SearchableDropdown.js
```diff
@@ -264,8 +274,15 @@ SearchableDropdown.propTypes = {
     inputId: isRequiredIf(string, props => props.label),
     /** Value of input field */
     inputValue: string,
-    /** value to be displayed in dropdown in case of search with no match */
-    noMatch: string.isRequired,
+    noMatch: oneOfType([
+        /** value to be displayed in dropdown in case of search with no match */
+        string.isRequired,
```
selbekk: I think you can remove `isRequired` here

selbekk (dismissed):
Looks good to me!

packages/ffe-searchable-dropdown-react/src/SearchableDropdown.js
patrisk: Then I'll clean up the proptypes and it should be good to go.

selbekk:
@patrisk that last commit needs to be prefixed with `fix` instead of `feat` - or just squashed into the original commit - before I'll be willing to merge.

The reason is that we use this "conventional commits" system to auto-generate changelogs - and "set proper proptypes" is not a new feature imo :)

patrisk:
@selbekk yeah, I realize the mistake. I squashed it so it's only the one commit now.

selbekk (approved):

selbekk:
Great job - thanks!

PR #426, opened 2018-09-26
fix(ffe-context-mesage): Removed dependency on React

42tte:
This version removes unneeded dependencies on react and react-dom.

### DIFF
--- a/packages/ffe-context-message/package.json
+++ b/packages/ffe-context-message/package.json
@@ -19,13 +19,10 @@
     "test": "npm run lint"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


kwltrs:
LGTM. But typo in the commit message: fix(ffe-context-me**ss**age). Needs to be corrected because it is used for generating the CHANGELOG.

42tte:
Får inte force push:at, gör om och skriver rätt :)

ivarni:
I think this might be a remnant of an old dev-server setup and if that is the case maybe you can remove the `test` npm script that tries to run eslint as well if there's no JS code to lint any more?

PR #430, opened 2018-09-26
fix(ffe-core): Visual indication on left menu

pethel:
Solves: https://github.com/SpareBank1/designsystem/issues/324

### DIFF
--- a/src/components/components-list/ComponentsListRenderer.js
+++ b/src/components/components-list/ComponentsListRenderer.js
@@ -13,31 +13,42 @@ export function ComponentsListRenderer({ items, useIsolatedLinks = false }) {
 
     return (
         <ul className="sb1ds-components-list">
-            {items.map(({ heading, name, slug, content }) => (
-                <li
-                    className={classNames(
-                        heading && 'sb1ds-components-list__item--heading',
-                        'sb1ds-components-list__item',
-                    )}
-                    key={name}
-                >
-                    <Link
-                        noUnderline={true}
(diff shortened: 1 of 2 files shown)


src/components/components-list/ComponentsListRenderer.js
```diff
@@ -13,31 +13,39 @@ export function ComponentsListRenderer({ items, useIsolatedLinks = false }) {
...
-                    className={classNames(
-                        heading && 'sb1ds-components-list__item--heading',
-                        'sb1ds-components-list__item',
-                    )}
-                    key={name}
-                >
-                    <Link
-                        noUnderline={true}
-                        className={classNames('sb1ds-components-list__link', {
-                            'sb1ds-components-list__link--heading': heading,
```
antidecaf, line 27: This line should be added back.

antidecaf (changes requested):
Thanks for the PR! Everything looks good, just needs a couple of minor changes:

* The commit message indicates a change in `ffe-core`, but there are no changes in that package. Thus, `fix(ffe-core):` should be rewritten to just `fix:`
* According to the description this PR solves #326, but the right issue number is #324.
* The `--heading` modifier has been removed from top level links in the sidebar - see separate comment. I think we need to keep that in place, even if the visual design is wrong (that can be fixed in a later PR when the look of the sidebar has been redesigned).

pethel:
Sorry about the confusion on issue number and commit message. Should be fixed now

antidecaf (approved):

PR #437, opened 2018-10-01
fix(ffe-header): set z-index on `.ffe-header` root

### LINKED ISSUE #436: The header could use a z-index

`ffe-header`

We have a massive background circle in one of apps that is basically used for creating a "rounded" divider line in one of our apps. We're using z-index to hide it on mobile screens.

The ffe-header does not come with a z-index and disappears behind this background.

I can't imagine anyone ever wanting the header to be hidden under other content on the sreen so I suggest giving it a z-index of **over 9000**.

Alternatively we can override the style locally if someone sees an issue with setting a z-index on the `.ffe-header` class.

ivarni:
fixes #436

### DIFF
--- a/packages/ffe-header/less/ffe-header.less
+++ b/packages/ffe-header/less/ffe-header.less
@@ -4,6 +4,7 @@
 
     background: @ffe-white;
     text-align: center;
+    z-index: 999;
 
     &__user-nav-list--visible,
     &__site-nav-list--visible {


selbekk (dismissed):
I think you need to specify that ffe-header is a positioned element (`position: relative`) for `z-index` to have any effect. Positioning it gives it a new stacking context. [This article](https://philipwalton.com/articles/what-no-one-told-you-about-z-index/) explains it pretty well.

Also check out the docs on mdn: https://developer.mozilla.org/en-US/docs/Web/CSS/z-index

ivarni:
I've tested it locally and the z-index was applied in Firefox, Chrome and IE. 

The other element that overlapped it is positioned and I don't really see the point in messing with the position of the `.ffe-header` to fix something that isn't broken.

antidecaf (commented):
`&__user-nav-list` and `&__site-nav-list` have z-index set to 999 already (line 420). Not sure if or how using the same value on the container will affect those elements, but it's worth double checking before merging.

selbekk:
if it works it's okay by me - but that just goes to prove z-index is hard to fathom :P

ivarni:
I usually test things before creating a PR and I can't see anything that looks off either with positioning or the user nav.

antidecaf (approved):

ivarni:
If we're very unsure about this I can override it locally. Maybe that's better than risking regressions?

antidecaf:
I'm not really that unsure. :P I've had a look at several header related bugs caused by overlapping elements lately and I think specifying a z-index on the root class is pretty much necessary at this point.

ivarni:
I just get unsure when I've tested it in 3 browser and then get 2 people telling me it won't work.

I guess we'll find out if it breaks anything.

Issue #438, opened 2018-10-01
ffe-formatters: minDecimals for formatPercentage


cecilieboe:
#### Which package is this issue related to?
ffe-formatters

#### Describe your issue (screenshots welcome!)
I want to use formatPercentage to show "1,00 %", but by now it is only possible to specify the maximum number of decimals, not the minimum number, so formatPercentage(1) gives "1 %".

#### What is the expected behavior? 
I would like to be able to get "1,00 %" as the formatted percentage.

#### What is the actual behavior?
It is not possible to get "1,00 %" as the output of formatPercentage.

#### Steps to reproduce
formatPercentage(1)

kwltrs:
Fixed by #439

PR #439, opened 2018-10-01
feat(ffe-formatters): Add minDecimals option to formatPercentage.

cecilieboe:

### DIFF
--- a/packages/ffe-formatters/src/formatPercentage.js
+++ b/packages/ffe-formatters/src/formatPercentage.js
@@ -1,6 +1,7 @@
 import { NON_BREAKING_SPACE } from './internal/unicode';
 
 const defaults = {
+    minDecimals: 0,
     maxDecimals: 2,
 };
 
@@ -10,7 +11,7 @@ export default function formatPercentage(amount, opts = defaults) {
     }
 
     const percentage = amount.toLocaleString('nb-NO', {
-        minimumFractionDigits: 0,
+        minimumFractionDigits: opts.minDecimals,
         maximumFractionDigits: opts.maxDecimals,
     });
     return `${percentage.replace('.', ',')}${NON_BREAKING_SPACE}%`;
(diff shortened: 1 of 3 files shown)


packages/ffe-formatters/src/formatPercentage.js
```diff
@@ -10,7 +11,7 @@ export default function formatPercentage(amount, opts = defaults) {
     }
 
     const percentage = amount.toLocaleString('nb-NO', {
-        minimumFractionDigits: 0,
+        minimumFractionDigits: opts.minDecimals,
```
ivarni, line 14: It probably does not have any significance but for people already passing an `opts` object with just `maxDecimals`, `minimumFractionDigits` are going to be passed as `undefined` since the default arg won't get used. AFAICT it's treated the same as `0` by `Number.toLocaleString` but in case there are edge-cases maybe we should be doing

```javascript
export default function formatPercentage(amount, opts = {}) {
    if (typeof amount !== 'number') {
        return '';
    }

    const percentage = amount.toLocaleString('nb-NO', {
        minimumFractionDigits: opts.minDecimals || defaults.minDecimals,
        maximumFractionDigits: opts.maxDecimals || defaults.maxDecimals,
    });
    return `${percentage.replace('.', ',')}${NON_BREAKING_SPACE}%`;
}
```

ivarni (approved):

ivarni:
Tips: Om du legger til teksten `Fixes #438` i bunn av commit-meldingen kobles PRen automatisk til issuet og issuet lukkes når PRen merges.

packages/ffe-formatters/src/formatPercentage.js
kwltrs: I agree on that this way of setting default values for `opts` is flawed and that we should fix it.
But also I approve and merge this pull request as this flaw was there from before.

cecilieboe:
Takk for tips og heads up @ivarni - får ikke fikset i denne omgang siden den allerede er merget, men skal ta det med neste gang :)

PR #445, opened 2018-10-03
Inline grid

### LINKED ISSUE #388: Grid vs spacing
[ffe-grid](https://github.com/SpareBank1/designsystem/tree/develop/packages/ffe-grid)[(-react)](https://github.com/SpareBank1/designsystem/tree/develop/packages/ffe-icons-react) comes with gutters, which makes it difficult to nest grids or use it in any other context than page layout. This has caused some issues in cases where it has been used for internal layout in components or sections. At the moment ffe-grid is the only alternative for creating layouts, and we need to figure out how to handle this in components.

One suggestion (originally intended to standardize spacing in general) is to remove all hardcoded spacing from the grid and other components, to be replaced by a set of predefined variables throughout the design system. In the case of ffe-grid, could removing the gutters altogether in favor of letting the contents of each column deal with spacing be a possible way forward?

Input wanted!

antidecaf:
### Suggested feature

This PR adds `InlineGrid`, a component that renders a `Grid`, but with no spacing. Stripping the spacing allows us to nest grids and create layouts not only for entire pages, but also for separate components, forms, content sections, etc.

#### Why do we need this?

Currently, the design system includes no tools to create layouts other than `ffe-grid(-react)`. While pages can be laid out nicely using the regular grid, it's not suited for other tasks due to internal spacing/gutters. Any lower level layouts are left up to the consumer to deal with, meaning there are many different implementations throughout our apps. Thus, an inline grid has potential to ease the process of writing markup and styling in apps considerably.

Another potential win is a more pragmatic approach to setting element widths. With no tool to use, consumers will typically pick a value specified in pixels or percentages that works on a case by case basis, but patterns established in similar pages/components are less likely to be considered. A grid implementation will limit the options a bit, while ensuring that different elements are generally aligned to one another.

Fixes #388

### DIFF
--- a/packages/ffe-grid-react/src/InlineGrid.js
+++ b/packages/ffe-grid-react/src/InlineGrid.js
@@ -1,0 +1,27 @@
+import React from 'react';
+import classNames from 'classnames';
+import { node, string } from 'prop-types';
+
+export default function InlineGrid(props) {
+    const { className, element: Element, ...rest } = props;
+
+    return (
+        <Element
+            className={classNames(className, 'ffe-grid', 'ffe-grid--inline')}
+            {...rest}
+        />
+    );
+}
+
+InlineGrid.propTypes = {
+    /** Any children of a InlineGrid must be a GridRow */
+    children: node,
+    /** Any extra classes are attached to the root node, in addition to ffe-grid classes */
+    className: string,
+    /** Specify the DOM element being used to create the Grid */
+    element: string,
+};
+
+InlineGrid.defaultProps = {
+    element: 'div',
+};
--- a/packages/ffe-grid-react/src/InlineGrid.md
+++ b/packages/ffe-grid-react/src/InlineGrid.md
@@ -1,0 +1,34 @@
+`InlineGrid` er et alternativ til `Grid`, men uten noen form for spacing i grid, rader eller kolonner. `<InlineGrid />` er i struktur identisk med `<Grid />` - den tar i mot de samme modifierne, bruker samme props (minus `condensed` og `topPadding`) og bygger rader og kolonner med `<GridRow />` og `<GridCol />`.
+
+`InlineGrid` kan nøstes, i motsetning til `Grid`. Med andre ord kan man ha flere grids inni hverandre. Denne komponenten er ideell til å lage layout på lavere nivå, som i komponenter, skjemaer, widgeter og lignende.
+
+```js
+<InlineGrid>
+    <GridRow>
+        <GridCol sm="6">Litt innhold til venstre</GridCol>
+        <GridCol sm="6">Litt innhold til høyre</GridCol>
+    </GridRow>
+</InlineGrid>
+```
+
+```js
+<InlineGrid>
+    <GridRow>
+        <GridCol sm="12" md="6" background="grey-warm">
+            Litt innhold til venstre
+        </GridCol>
+        <GridCol sm="12" md="6">
+            <InlineGrid>
+                <GridRow>
+                    <GridCol sm="6" background="blue-ice">
+                        Grid inni grid - venstre
+                    </GridCol>
+                    <GridCol sm="6" background="green-mint">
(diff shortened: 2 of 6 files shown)


packages/ffe-grid-react/src/InlineGrid.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import classNames from 'classnames';
+import { node, string } from 'prop-types';
+
+export default function InlineGrid(props) {
+    const { children, className, element, ...rest } = props;
+
+    const Element = element || 'div';
```
ivarni: Minor, men jeg syns egentlig det er penere å bruke `InlineGrid.defaultProps` for defaults.

packages/ffe-grid-react/src/InlineGrid.md
```diff
@@ -0,0 +1,34 @@
+`InlineGrid` er et alternativ til `Grid`, uten intern spacing. `<InlineGrid />` er i struktur identisk med `<Grid />` - den tar i mot de samme modifierne, bruker samme props (minus `condensed` og `topPadding`) og bygger rader og kolonner med `<GridRow />` og `<GridCol />`.
```
ivarni: Er det bare gutters/intern spacing du fjerner eller fjerner du også paddingen på kantene?

ivarni (commented):
Ser ut som et greit kompromiss å tilby begge deler 👍

packages/ffe-grid-react/src/InlineGrid.md
antidecaf: Fjerner i utgangspunktet all spacing som følger med by default.

ivarni: Kanskje vi heller skal si det da :) Med "intern spacing" tolker jeg det som at det er gutterne mellom kolonnene som fjernes og ikke paddingen på utsiden.

packages/ffe-grid-react/src/InlineGrid.js
```diff
@@ -0,0 +1,27 @@
...
+
+    return (
+        <Element
+            className={classNames(className, 'ffe-grid', 'ffe-grid--inline')}
+            {...rest}
+        >
+            {children}
```
selbekk: Nitpick: You don't need to deconstruct children here, you can just pass them on. That simplifies things a bit:

```js
export default function InlineGrid(props) {
  const { className, element: Element, ...rest } = props;
  return <Element className={...} {...rest} />;
};
```

selbekk (dismissed):
Very nice!

Just a suggestion - but instead of expanding the API, could we just use the cascade?

In other words:

```css
.ffe-grid .ffe-grid,
.ffe-grid .ffe-grid .ffe-grid--row,
.ffe-grid .ffe-grid .ffe-grid--col { 
  margin: 0; 
  padding: 0; 
}
```

This way, we _can_ nest grids without this leading to any trouble.

If not - the code as is is also fine.

packages/ffe-grid-react/src/InlineGrid.md
antidecaf: Ja, det kan sikkert ordlegges litt annerledes. Skal fikse!

antidecaf:
@selbekk Yeah, that would probably be a good solution too. But I'm thinking it might add to the perceived complexity of the grid system and perhaps feel a bit unpredictable if rows and columns behave differently depending on nesting? I also think extending the API could be a good way to make a point out of adding a dedicated option to create lower level layouts, although in the end it means a bit more code.

ivarni:
Keep things explicit. And this way people can use the no-gutters version even if the page itself is not layed out with a `<Grid>`. If we went for nesting you would need an outer grid to get the inline one.

selbekk:
Good points both of you. Fair enough - keep as is then :)

ivarni (approved):

hdaljord (approved):
Looks good.

Issue #415, opened 2018-09-21
Peer dependencies are not maintained (need update for ffe-core@14)


henrikhermansen:
#### Which package is this issue related to?
Many of our packages

#### Describe your issue (screenshots welcome!)
I have just now installed all the newest versions of ffe-packages,
but peer dependencies are not updated for ffe-core@14.
All of the following warnings are just noise - all packages works fine with ffe-core@14:

```
npm WARN @sb1/ffe-buttons@8.0.7 requires a peer of @sb1/ffe-core@^12.0.0 || ^13.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @sb1/ffe-buttons-react@11.1.7 requires a peer of @sb1/ffe-buttons@^7.0.2 but none is installed. You must install peer dependencies yourself.
npm WARN @sb1/ffe-context-message@3.0.18 requires a peer of @sb1/ffe-core@^12.0.0 || ^13.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @sb1/ffe-core-react@3.2.2 requires a peer of @sb1/ffe-core@^12.0.0 || ^13.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @sb1/ffe-datepicker@5.0.32 requires a peer of @sb1/ffe-core@^12.0.0 || ^13.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @sb1/ffe-form-react@3.2.3 requires a peer of @sb1/ffe-core@^12.0.0 || ^13.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @sb1/ffe-grid@10.0.4 requires a peer of @sb1/ffe-core@^12.0.0 || ^13.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @sb1/ffe-grid-react@9.0.2 requires a peer of @sb1/ffe-grid@^7.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @sb1/ffe-layout-legacy@1.0.2 requires a peer of @sb1/ffe-core@^12.2.0 || ^13.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @sb1/ffe-lists@5.0.17 requires a peer of @sb1/ffe-core@^12.0.0 || ^13.0.0 but none is installed. You must install peer dependencies yourself.
[...]

PR #447, opened 2018-10-03
Add section about peer-dependencies

ivarni:
I was thinking something like this can remedy the issue in #415. The only problem is... The depended URL does not show peerDepedencies and `ffe-form` is showing up because it has an optionalDependency so for this to work we'd need to use both.

Optional means that npm will try to install it but the process won't fail if it can't. I'm not sure what exactly happens if the project already depends on a different version of said package, that's maybe something we should test.

So in short, this can't be merged yet, we need to figure out a way to detect packages that will need updates. I just figured I'd get the ball rolling on this.

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -12,4 +12,14 @@ document. TL;DR:
 - Follow our code standards
 - Write tests where applicable
 - Be courteous and respect our code of conduct
+
+*NOTE
+
+If you are making any breaking changes that will lead to a new major release, please
+visit the package's dependents and update their peerDependencies (remember to test
+that they still work)
+
+You can find dependents by visiting https://www.npmjs.com/browse/depended/@sb1/<package>.
+(e.g: https://www.npmjs.com/browse/depended/@sb1/ffe-core)
+
 -->


selbekk (approved):
Do we do this before or after we release the new breaking version?

ivarni:
There's a new breaking version?

selbekk:
> If you are making any breaking changes that will lead to a new major release, please
> visit the package's dependents and update their peerDependencies

Do we follow your instructions in the pull request that leads to the breaking change, or directly after?

ivarni:
In the same PR so it all gets released at the same time?

ivarni:
Shouldn't matter what order the commits are in, if that's what you meant.

selbekk:
As long as our release process doesn't crash because we're referencing versions that aren't available at the time of release, that sounds fine with me.

ivarni:
I'll have to defer to the people who know how the release process works to answer that.

antidecaf (approved):

ivarni:
@selbekk why on earth would you merge this?

> So in short, this can't be merged yet, we need to figure out a way to detect packages that will need updates. I just figured I'd get the ball rolling on this.

This isn't adressed yet.

> As long as our release process doesn't crash because we're referencing versions that aren't available at the time of release, that sounds fine with me.

It does.

selbekk:
Oh, fair enough. I took @antidecaf 's approval as an OK that it didn't indeed crash. It's been reverted.

ivarni:
Well, with the PR closed at least we can leave this behind us. We still have an unsolved issue with peer dependencies though.

PR #449, opened 2018-10-04
Visible underline on tertiary button in inline mode

### LINKED ISSUE #446: Tertiary button has no underline in inline button group

ffe-buttons

Tertiary button lacks underline when it is used in an inline button group:

image

I expect tertiary button to always have an underline, as it has in all other button groups. 

image

No underline when placed inside an inline button group.
#### Steps to reproduce

This can be seen in the design guide at https://design.sparebank1.no/styleguidist/index.html#!/ButtonGroup, scroll down to Inline

hdaljord:
**Bug fix**

This PR fixes a bug where the underline of tertiary button is not visible in inline mode.

This fix makes all tertiary buttons align with the sketch file design of tertiary button:

After fix:

image

Fixes #446.

### DIFF
--- a/packages/ffe-buttons/less/inline-base-button.less
+++ b/packages/ffe-buttons/less/inline-base-button.less
@@ -74,6 +74,7 @@
     &::after {
         border-bottom: solid 1px currentColor;
         content: ' ';
+        display: inherit;
         width: 100%;
     }
 


packages/ffe-buttons/less/inline-base-button.less
```diff
@@ -74,6 +74,7 @@
     &::after {
         border-bottom: solid 1px currentColor;
         content: ' ';
+        display: inherit;
```
benmatthews, line 77: I think maybe this should be `block` and not `inherit` actually... despite my previous comment ;-)

antidecaf (approved):

packages/ffe-buttons/less/inline-base-button.less
antidecaf: You're probably right, what value it inherits isn't exactly given, although in the styleguide examples it's block. I may have been a bit trigger happy on merging this. :fearful:

hdaljord: I agree, should have seen it myself 😦

PR #452, opened 2018-10-06
fix(ffe-account-selector-react): Suggestions will now correctly lose …

### LINKED ISSUE #435: AccountSelectorMulti won't close when focus is moved to another element
ffe-account-selector-react

When opening the account selector using the button on the right, the account selector won't close when focus is moved to another element in the page.
image

The account-selector should close when focus is moved to another element in the page.

The account-selector stays open even though focus is moved to a different element.

#### Steps to reproduce
- Click the button/arrow to open the account selector and see that the list of accounts is opened and the textfield does not have focus
- Click outside the account selector and notice that the list of accounts is still open even though focus is not in the account selector

simjes:
…focus

The suggestion list will now close when AccountSelector lose focus after opening it by clicking the chevron

Fixes #435

### DIFF
--- a/packages/ffe-account-selector-react/src/subcomponents/input-field/InputField.js
+++ b/packages/ffe-account-selector-react/src/subcomponents/input-field/InputField.js
@@ -42,7 +42,14 @@ class Input extends Component {
 
     onExpandOrCollapse(e) {
         e.preventDefault();
-        this.props.onExpandOrCollapseClick();
+
+        const { isSuggestionsShowing } = this.props;
+        if (isSuggestionsShowing) {
+            this.onBlur();
+        } else {
+            e.currentTarget.previousElementSibling.focus();
+            this.onFocus();
+        }
     }
 
     componentWillReceiveProps(nextProps) {
@@ -65,7 +72,7 @@ class Input extends Component {
             suggestionListId,
             name,
             readOnly,
-            locale
+            locale,
         } = this.props;
 
         const showReset = !readOnly && value.length > 0;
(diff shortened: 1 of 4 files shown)


packages/ffe-account-selector-react/src/subcomponents/input-field/InputField.js
```diff
@@ -40,9 +40,14 @@ class Input extends Component {
         this.props.onReset();
     }
 
-    onExpandOrCollapse(e) {
+    onExpandOrCollapse(e, isSuggestionsShowing) {
```
selbekk: Can you get this value from `this.props.isSuggestionsShowing` instead of passing it as an argument?

selbekk (changes requested):
Great job on this PR - especially with adding tests and what not.

I would love for the callback to refer to `this.props` instead of passing the value as an argument to the event handler - but that's the only change I can think of. Good stuff!

simjes:
@selbekk Updated :)

selbekk (approved):

selbekk:
So can you please explain to me why you're doing this instead of what we did previously? I'm not well enough versed with this code to understand its implications.

antidecaf (approved):

simjes:
e.preventDefault() will prevent focus being set on an element so the onBlur() would not fire when clicking outside the inputField component.
Instead I put focus on the input element when you click the button. Setting focus on the button itself would also work, but there would be a border on the button and I didn’t want to change the style.

I can’t remember what it was, but there was another issue with putting focus on the button.

selbekk:
Thanks for the explanation @simjes :+1: and thank you for your first contribution!

PR #456, opened 2018-10-09
fix(ffe-lists): added margin collapse to numbered list and bullet list

pethel:
Solves https://github.com/SpareBank1/designsystem/issues/290 if you wanted it solved

### DIFF
--- a/packages/ffe-lists/less/regular-lists.less
+++ b/packages/ffe-lists/less/regular-lists.less
@@ -52,7 +52,7 @@
 .ffe-bullet-list {
     padding: 0;
     text-align: left;
-    margin: 0 0 20px;
+    margin: -10px 0 20px;
 
     > li {
         margin-top: 10px;


packages/ffe-lists/less/regular-lists.less
```diff
@@ -52,7 +52,7 @@
 .ffe-bullet-list {
     padding: 0;
     text-align: left;
-    margin: 0 0 20px;
+    margin: -10px 0 20px;
```
selbekk, line 55: Instead of inverting the margin here, how about just removing the padding from the first list item?

In other words:

```css
> li {
  margin-top: 10px;
  &:first-of-type {
    margin-top: 0;
  } 
  // ...
}
```

Both obviously work, but it's easier (at least to me) to comprehend how the `:first-of-type` selector works, than negative margins.

henrikhermansen:
Closing stale PR on behalf of @pethel

PR #464, opened 2018-10-19
Dark variant

lotorvik:
New dark variant (dark={true}) for internal applications as earlier discussed. See changelog for the affected components. We have added examples in the documentations for the main components that have the new option.

Three new colors added as agreed with --dark as suffix to name. These are not a part of the regular ffe palett and should only be used with --dark components.

Also added --condensed to RadioSwith

Edit: added an example of a dark component.
textarea


packages/ffe-form/less/tooltip.less
```diff
@@ -62,4 +62,34 @@
...
+
+    &--dark {
+        &__icon {
+            background-color: transparent;
+            border: 2px solid @ffe-grey-silver;
+            color: @ffe-blue-focus;
+
+            &:hover {
+                border-color: @ffe-white;
+                //color: @ffe-grey-gunmetal--dark;
```
antidecaf: Any reason this is commented out? Same thing on line 89.

lotorvik: My bad. We did some last minute testing to get rid of a color and forgot to remove the comment with the old color. I'll fix it right away.

selbekk:
Fantastic effort! This seems like the culmination of a lot of work. 

Remembering to pass `dark` to a ton of components while implementing internal applications seems like a lot of work. Could we consider using React's context API to pass the `dark` value down to all components that are interested in it?

Something like:

```js
const Theme = React.createContext();
const ThemeProvider = props => (
  <Theme.Provider value={props.theme === 'dark'}>
    {props.children}
  </Theme.Provider>
);
const withTheme = Component => props => (
  <Theme.Consumer>
    {isDarkTheme => <Component {...props} isDarkTheme={isDarkTheme} />}
  </Theme.Consumer>
```

Then you can wrap your application (or themed part of your application) in a `<ThemeProvider />` and all themeable components with `withTheme`.

Anyhow, something for a future PR in any case.

selbekk:
Fantastic effort! This seems like the culmination of a lot of work. 

Remembering to pass `dark` to a ton of components while implementing internal applications seems like a lot of work. Could we consider using React's context API to pass the `dark` value down to all components that are interested in it?

Something like:

```js
const Theme = React.createContext();
const ThemeProvider = props => (
  <Theme.Provider value={props.theme === 'dark'}>
    {props.children}
  </Theme.Provider>
);
const withTheme = Component => props => (
  <Theme.Consumer>
    {isDarkTheme => <Component {...props} isDarkTheme={isDarkTheme} />}
  </Theme.Consumer>
```

Then you can wrap your application (or themed part of your application) in a `<ThemeProvider />` and all themeable components with `withTheme`.

Anyhow, something for a future PR in any case.

selbekk:
Fantastic effort! This seems like the culmination of a lot of work. 

Remembering to pass `dark` to a ton of components while implementing internal applications seems like a lot of work. Could we consider using React's context API to pass the `dark` value down to all components that are interested in it?

Something like:

```js
const Theme = React.createContext();
const ThemeProvider = props => (
  <Theme.Provider value={props.theme === 'dark'}>
    {props.children}
  </Theme.Provider>
);
const withTheme = Component => props => (
  <Theme.Consumer>
    {isDarkTheme => <Component {...props} isDarkTheme={isDarkTheme} />}
  </Theme.Consumer>
```

Then you can wrap your application (or themed part of your application) in a `<ThemeProvider />` and all themeable components with `withTheme`.

Anyhow, something for a future PR in any case.

selbekk:
Fantastic effort! This seems like the culmination of a lot of work. 

Remembering to pass `dark` to a ton of components while implementing internal applications seems like a lot of work. Could we consider using React's context API to pass the `dark` value down to all components that are interested in it?
Anyhow, something for a future PR in any case.

selbekk (commented):
Fantastic effort! This seems like the culmination of a lot of work. 

Remembering to pass `dark` to a ton of components while implementing internal applications seems like a lot of work. Could we consider using React's context API to pass the `dark` value down to all components that are interested in it?

Also - I'm concerned about the amount of bloat this adds to our customer-facing applications - especially when it comes to the CSS. Are we okay with this, or should we split all "dark theme" styles into their own LESS-files, that in turn needs to be included?

Again, a great effort on your part - I just have a few questions before we kick it off ;-)

antidecaf (dismissed):
This looks great.

I like @selbekk's `<ThemeProvider>` suggestion, but agree that it can wait for a future PR. I'm not really worried about bloating the CSS given the amount of code in question, but as that might change in the future, splitting the dark styles out into separate files does make sense.

ksawerysb1:
Wouldn't adding a theme="dark" instead of a dark=true make this more useful for customizing look and feel in the future?

ksawerysb1:
Wouldn't adding a theme="dark" instead of a dark=true make this more useful for customizing look and feel in the future?

ksawerysb1:
Wouldn't adding a theme="dark" instead of a dark=true make this more useful for customizing look and feel in the future?

ksawerysb1:
Wouldn't adding a theme="dark" instead of a dark=true make this more useful for customizing look and feel in the future?

packages/ffe-buttons-react/src/ActionButton.js
```diff
@@ -45,4 +45,6 @@ ActionButton.propTypes = {
     leftIcon: node,
     /** Icon shown to the right of the label */
     rightIcon: node,
+    /** Dark variant */
+    dark: bool,
```
ksawerysb1, line 50: Wouldn't adding a theme="dark" instead of a dark=true make this more useful for customizing look and feel in the future?

lotorvik:
Regarding splitting out --dark css in seperate files. I see the point about bloat, but seperating it also makes it harder to maintain. If somone decide to do a change in regular component he will for sure forget that it might affect the dark component if the style is not in same file. At the moment I feel the risk for --dark being "forgotten" is bigger then risk for bloat. What do you guys think?

The ThemeProvider idea sounds like a good idea for some projects. For our projects it would not work since we have --dark on many components. But we also have white cards inside the dark theme so on those we need to use the regular components without the --dark flag.

lotorvik:
Regarding splitting out --dark css in seperate files. I see the point about bloat, but seperating it also makes it harder to maintain. If somone decide to do a change in regular component he will for sure forget that it might affect the dark component if the style is not in same file. At the moment I feel the risk for --dark being "forgotten" is bigger then risk for bloat. What do you guys think?

The ThemeProvider idea sounds like a good idea for some projects. For our projects it would not work since we have --dark on many components. But we also have white cards inside the dark theme so on those we need to use the regular components without the --dark flag.

packages/ffe-buttons-react/src/BaseButton.js
```diff
@@ -87,6 +89,8 @@ BaseButton.propTypes = {
     leftIcon: node,
     /** Icon shown to the right of the label */
     rightIcon: node,
+    /** Dark variant */
+    dark: bool,
```
kwltrs, line 93: The `dark` props should have a default value of `false`, even though they are falsy by default. Our linting requires strict equals comparisons and checks like `props.dark === false` would then behave unexpectedly.

lotorvik: To be honest I started doing that but since most components did not use default variables for bool I removed it againt :-)

lotorvik: As agreed on in the designforum meeting I have added default props all places that accept the dark property

packages/ffe-buttons-react/src/ActionButton.js
lotorvik: In todays designforum meeting we agreed to keep the variabel dark at this time. But we are aware this will result in a breaking change if we decide to add multiple theme support in the future.

selbekk (approved):

selbekk:
> For our projects it would not work since we have --dark on many components. But we also have white cards inside the dark theme so on those we need to use the regular components without the --dark flag.

This is the scenario that I'm the most concerned about. Are the light components designed to work in a dark context? If you want a button inside a light component surrounded by dark background - should that have the `dark` flag set?

antidecaf (approved):

lotorvik:
@selbekk I would say that is up to the designer. But based on how the components are build I would guess it is more natural to have light blocks/cards in a dark design. Then having dark blocks/card in a light design.  But I'm no designer, I just do as the designers tell me. 

In all the documentation examples for dark components, it is mentioned that they only should be used in internal applications at SB1.

kwltrs (approved):

Issue #372, opened 2018-08-16
Future-proof React 17 components


selbekk:
#### Which package is this issue related to?
A subset of our React packages:
- `ffe-account-selector-react`
- `ffe-datepicker-react`
- `ffe-decorators-react`
- `ffe-tables-react`

#### Describe your issue (screenshots welcome!)
In React 17, a few lifecycle methods (`componentWillMount`, `componentWillReceiveProps`, and `componentWillUpdate`) will be deprecated and removed. 

In order to simplify this coming update, we should rewrite all components so that they use the modern alternatives. 

To ensure that people still running older versions of React still can use our component library, any usage of [`static getDerivedStateFromProps`](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) and [`getSnapshotBeforeUpdate`](https://reactjs.org/docs/react-component.html#getsnapshotbeforeupdate) must lead to polyfilling the component in question with the `react-lifecycle-compat` polyfill. You can find it here: https://github.com/reactjs/react-lifecycles-compat

ivarni:
deprecated **and** removed? 😱

selbekk:
Well, they are renamed to `UNSAFE_componentWill*`, so you can still use them for a while longer if you don't have the time to rewrite.

kwltrs:
This is road map material. If there is no other reason for changing the lifecycle callbacks (and thus forcing consumers to upgrade react or polyfilling stuff) than avoiding deprecation warnings, I feel we should wait until an upgrade would make it more worth it.

selbekk:
Just for reference - the polyfill is 916 bytes gzipped, and will most likely be included in all large libraries (react-redux, react-router etc) for the foreseeable future. In other words - there isn't a real cost connected to this.

Since this imo needs to be done at some point - what's the reasoning for not doing it now?

kwltrs:
The cost isn't the file size of the polyfill, but the cost of development when having to upgrade or polyfill out of the blue.

selbekk:
That cost is fully contained within FFE - the polyfill is applied in FFE code only, and will make our components backwards compatible, delaying the cost of upgrading as well.

ivarni:
As far as the deprecated lifecycle callbacks are concerned, only the *UNSAFE_* prefixed names will work in 17.x but for anyone using very old versions those names will not work so using them would potentially be a breaking change. It seems better to rewrite the components that use these to instead use one of the alternatives.

PR #466, opened 2018-10-20
feat(ffe-tables-react): Migrate from deprecated componentWillReceiveP…

simjes:
This is a suggested solution for #372 for ffe-tables-react. I am not sure if this is the best way to solve it, especially  for the SortableTable, see: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html. “If you’re updating derived state unconditionally or updating it whenever props and state don’t match, your component likely resets its state too frequently”.

I am open for suggestions for how to improve this solution.

### DIFF
--- a/packages/ffe-tables-react/src/SortableTable/SortableTable.js
+++ b/packages/ffe-tables-react/src/SortableTable/SortableTable.js
@@ -1,44 +1,31 @@
-import React, { Component } from 'react';
-import PropTypes from 'prop-types';
+import PilNedIkon from '@sb1/ffe-icons-react/lib/pil-ned-ikon';
 import classNames from 'classnames';
 import equal from 'deep-equal';
-import PilNedIkon from '@sb1/ffe-icons-react/lib/pil-ned-ikon';
-import sortData from './sort-data';
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
 import Table from '../Table';
+import sortData from './sort-data';
+import memoize from 'memoize-one';
 
 class SortableTable extends Component {
     constructor(props) {
         super(props);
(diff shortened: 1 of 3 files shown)


selbekk (approved):

henrikhermansen (dismissed):
This component SortableTable is quite complex. I see we do have a shouldComponentUpdate, so I guess using derived state isn't too bad here.

This commit will just party fix #372 as it only changes ffe-tabls-react, right? Or has the other packages been handled already?

simjes:
Yes, this is only a partial fix for #372.  #369 covered one of the other ones, but `ffe-datepicker-react` and `ffe-decorators-react` are still missing

packages/ffe-tables-react/src/SortableTable/SortableTable.js
```diff
@@ -37,8 +32,18 @@ class SortableTable extends Component {
                           prevState.descending,
                       )
                     : nextProps.data,
-            }));
+                props: nextProps,
```
selbekk: This is a pragmatic way of doing things, but I think we can do even better. Saving all the previous props in state seems like an anti-pattern we can avoid.

Perhaps we can memoize the `sortData` method, and call it on every render? This way we can omit keeping the sorted table data in state as well! 

You can read more about this approach in the ["You might not need derived state"](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization) blog post by the React team.

selbekk (changes requested):

simjes:
Updated @selbekk

selbekk (approved):
That looks god damn amazing. :tada: :100:

Issue #465, opened 2018-10-19
Husky error


lotorvik:
#### Which package is this issue related to?
husky

#### Describe your issue (screenshots welcome!)
I tried to do a regular comit but got the following error message:
Error: Received $GIT_PARAMS as value for -e | --edit, but GIT_PARAMS is not available globally.

#### What is the expected behavior? 
I expcted to be able to do a comit to GitHub

#### What is the actual behavior?
I can't comit to GitHub

#### Steps to reproduce
This error message came after I updated from develop branch and got some changes in the husky configuration late september.

#### Resolution
If you take a look at https://github.com/typicode/husky under "Upgrading from 0.14" it seems we are using an old syntax? When I changed it to the new syntax everything worked fine. I should mention that I'm working on Windows so this might not be a problem for everyone...

Change from:
commitlint -e $GIT_PARAMS
to:
commitlint -E HUSKY_GIT_PARAMS

I'm not familiar with husky but this solves the problem for me and since this is documented I guess it should be fine to do this. I can do a PR for this if somone could test this on linux/mac?

simjes:
+1 for this, had to use Bash for Windows to commit

ivarni:
Seems to work fine here with the change described above.

Tested on *Ubuntu 16.04.5 LTS*

output
```bash
➜  designsystem git:(foobar) ✗ git diff | grep commitlint
-            "commit-msg": "commitlint -e $GIT_PARAMS",
+            "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
➜  designsystem git:(foobar) ✗ git commit -m "fix(ffe-core): Test commitlint change" .
husky > pre-commit (node v8.11.3)
🔍  Finding changed files since git revision cb262fa.
🎯  Found 0 changed files.
✅  Everything is awesome!
husky > commit-msg (node v8.11.3)
⧗   input: fix(ffe-core): Test commitlint change
✔   found 0 problems, 0 warnings

[foobar 6039b66] fix(ffe-core): Test commitlint change
 1 file changed, 1 insertion(+), 1 deletion(-)
```

PR #467, opened 2018-10-22
Fix husky configuration

lotorvik:
Resolves the case #465 so we can comit again from Windows.


ivarni (approved):
Worked fine on Ubuntu 16.04

kwltrs (approved):

PR #468, opened 2018-10-22
fix(ffe-grid-react): remove nested grid warnings and tests

antidecaf:
Since `InlineGrid` was implemented in #445, nesting of grids is now accepted. This PR removes related warnings and tests that have thus become obsolete.

### DIFF
--- a/packages/ffe-grid-react/src/Grid.js
+++ b/packages/ffe-grid-react/src/Grid.js
@@ -1,43 +1,32 @@
-import React, { Component } from 'react';
+import React from 'react';
 import classNames from 'classnames';
 import { bool, node, string } from 'prop-types';
 
-import { checkForNestedComponent } from './utils';
+export default function Grid(props) {
+    const {
+        children,
+        className,
+        condensed,
+        element,
+        topPadding,
+        ...rest
+    } = props;
 
-export default class Grid extends Component {
-    componentDidMount() {
-        /* istanbul ignore else: there is no else  */
-        if (process.env.NODE_ENV !== 'production') {
-            checkForNestedComponent(this.props.children, Grid);
-        }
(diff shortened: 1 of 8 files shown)


selbekk:
Shouldn't we still check for the nesting of regular grids?

ivarni:
I was under the impression we only allowed nesting the inline version, you'd still get unexpected results if you nest the ones with gutters.

ivarni:
Also, you still want to avoid having a `GridCol` as a descendant of another `GridCol` unless that inner `GridCol` is within an `InlineGrid` (with a `GridRow`).

```xml
<Grid>
    <GridRow>
        <GridCol>
            <GridCol>
                derp
            </GridCol>
        </GridCol>
    <GridRow>
</Grid>
```

Should still not be allowed.

If anything, the check should possible be expanded to making sure 

* any `GridCol` is contained directly within a `GridRow`
* if a `GridCol` is in the subtree of another `GridCol` it must also be in the subtree of an `InlineGrid`
* any `GridRow` is contained directly within a `Grid` or `InlineGrid`
* If a `GridRow` is in the subtree of another `GridRow` it must also be in the subtree of an `InlineGrid`
* a `Grid` can never be in the subtree of an `InlineGrid`
* any `InlineGrid` should be in the subtree of a `Grid` (?)

I probably should have pointed this out when the inline stuff was introduced but I kinda forgot about these tests and since I didn't check out and run the branch in that PR I never saw any warnings.

kwltrs (approved):
`checkForNestedComponent` is a nice demo of parts of React's API. But I'm fine with removing it.

ivarni:
Is warnings being printed when using inline-grid still an issue or can this be closed?

PR #473, opened 2018-10-25
feat(ffe-lists): Add new modifier ffe-check-list__item--cross

### LINKED ISSUE #471: X instead of check support in CheckList

ffe-lists

I am using `ffe-check-list` to show a list of features for a given product - some of which are included and some that are not. I want to use our `KryssIkon` icon in `@ffe-red` for noting when something is not.

I've created a local override, which looks like this:

screenshot of how the not included checkmark would look like

I've implemented it with a class on the `li` element (`ffe-check-list__item--nope`), but I'm open for suggestions on how it could be solved.

selbekk:
This commit adds a new modifier for ffe-check-list list items to show that the item is not included in
whatever the list is about.

Resolves #471 

Not sure about the class name, nor how this should be implemented for the React component. We don't have any `ListItem` components I could add a flag to - should we perhaps add that?

### DIFF
--- a/packages/ffe-lists/less/regular-lists.less
+++ b/packages/ffe-lists/less/regular-lists.less
@@ -2,9 +2,9 @@
 //
 // Markup:
 // <ul class="ffe-bullet-list ffe-bullet-list{{modifier_class}}">
-//     <li>Bilforsikring</li>
-//     <li>Reiseforsikring</li>
-//     <li>Innboforsikring</li>
+//     <li class="ffe-bullet-list__item">Bilforsikring</li>
+//     <li class="ffe-bullet-list__item">Reiseforsikring</li>
+//     <li class="ffe-bullet-list__item">Innboforsikring</li>
 // </ul>
 //
 // --condensed  - Condensed
@@ -15,9 +15,10 @@
 //
 // Markup:
 // <ul class="ffe-check-list ffe-check-list{{modifier_class}}">
-//     <li>Massevis av ferdige, testede komponenter</li>
-//     <li>Likt design som resten av SpareBank 1</li>
-//     <li>Høyere utviklingshastighet</li>
+//     <li class="ffe-check-list__item">Massevis av ferdige, testede komponenter</li>
+//     <li class="ffe-check-list__item">Likt design som resten av SpareBank 1</li>
+//     <li class="ffe-check-list__item">Høyere utviklingshastighet</li>
+//     <li class="ffe-check-list__item ffe-check-list__item--cross">Null bugs</li>
 // </ul>
 //
 // --bg-sand    - Sand background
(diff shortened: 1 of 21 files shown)


ivarni:
The react component could take children as either `<li>`s or  `<ListItem>`s or you could just remove support for raw `li` elements. It shouldn't be a very difficult migration for consumers.

packages/ffe-lists/less/regular-lists.less
```diff
@@ -138,6 +141,14 @@
...
+    > li.ffe-check-list__item--cross {
+        &::before {
+            background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 200 200'%3E%3Cpath fill='@{ffe-red-fill}' d='M18.77,0c-3.562,0-7.125,1.347-9.834,4.044l-4.429,4.439c-5.381,5.394-5.381,14.25,0,19.68l71.19,71.83-71.66,71.83c-5.381,5.394-5.381,14.25,0,19.68l4.429,4.439c5.381,5.394,14.22,5.394,19.64,0l71.66-71.83,71.65,71.83c5.381,5.394,14.23,5.394,19.65,0l4.429-4.439c5.381-5.394,5.381-14.25,0-19.68l-71.19-71.83,71.66-71.83c5.381-4.914,5.381-13.78,0-19.21l-4.429-4.423c-5.381-5.394-14.22-5.394-19.64,0l-72.13,71.34-71.17-71.83c-2.691-2.697-6.257-4.044-9.819-4.044z'/%3E%3C/svg%3E");
+            color: @ffe-red;
+            content: '\2717';
+        }
+    }
+
     &--bg-sand {
         > li {
```
ivarni: Maybe this should be changed from styling the raw element and have a `.ffe-check-list__item` class instead? Would also mean not having to have a modifier for a non-existent class below, pluss we moved away from styling on raw elements somewhere around `v4` or something like that?

packages/ffe-lists/less/regular-lists.less
```diff
@@ -18,6 +18,7 @@
 //     <li>Massevis av ferdige, testede komponenter</li>
 //     <li>Likt design som resten av SpareBank 1</li>
 //     <li>Høyere utviklingshastighet</li>
+//     <li class="ffe-check-list__item--cross">Null bugs</li>
```
kwltrs: I miss the base class `ffe-check-list__item`. I know it doesn't do anything, but that's why it is called a modifier in BEM.

selbekk: yeah I agree, I'll implement what @ivarni suggested above

selbekk:
So this was the biggest scope creep I've done in a while. :slightly_frowning_face: 

Note that `ffe-lists-react` won't get a major bump, because it still works as intended with `ffe-lists@^5.0.0`  AND `6.0.0`. The big difference is that with `6.0.0`, you'd have to manually add the classes to the list items and definition term/descriptions.

ivarni:
Just wondering if those commits to the same package should be squashed?

selbekk:
The three first ones? Meh

Edit: I don't mind either way - but I don't see a huge downside to keeping them separated

ivarni:
So that won't lead to three releases? How will the generated CHANGELOG look?

My god won't someone think of the changelog

selbekk:
Relax, there's only one release :laughing: or well, 2, ffe-lists@major and ffe-lists-react@minor

The changelog will hopefully look pretty nice - with nice migration guides.

selbekk:
To move this forward, I went ahead and squashed the commits per @ivarni's request.

Please let me know if there are further changes required before this PR can be approved.

ivarni (approved):

antidecaf (approved):

PR #483, opened 2018-10-31
Two column lists

### LINKED ISSUE #472: Two-column support for check-lists
ffe-lists

When working with lists of product features, the lists can become pretty long and heavy on the left side:

A screenshot of a long list of items

I'm suggestion that for larger screens (medium and up for example), we can support a two-column layout:

image

It's impossible to do this with our current grid system, because we can't really apply it to a semantic list of items. We'd have to split it into two, and the order would get all messed up.

What do you think? I'd be happy to submit a PR if this is something we want.

selbekk:
This pull request adds support for two-column check lists.

End result looks like this: 

image

For the React component, I went for adding a `columns="2"` prop instead of a `twoColumns={true}` flag, just in case we'll want to add support for three columns later. 

Note that you can write `columns={2}` and `columns="2"` - that I added as a convenience for developers.

I first tried implementing the two-column layout by using [CSS Columns](https://developer.mozilla.org/en-US/docs/Web/CSS/columns), but that behaved a bit weirdly with wrapping margins and what not. Here's how that looked like:

image

Thus, I ended up with wrapping flexbox items and setting an explicit width of 40 %.

---

This branch will be merged into develop eventually, but since #471 is still open, I decided to branch out from that branch (since there was a lot of changes to the ffe-lists package there). I will change the base of this PR once that is merged.

Fixes #472

### DIFF
--- a/packages/ffe-lists-react/src/CheckList.js
+++ b/packages/ffe-lists-react/src/CheckList.js
@@ -1,9 +1,16 @@
 import React from 'react';
 import classNames from 'classnames';
-import { node, string } from 'prop-types';
+import { node, oneOf, string } from 'prop-types';
 
-const CheckList = ({ className, ...rest }) => (
-    <ul className={classNames('ffe-check-list', className)} {...rest} />
+const CheckList = ({ className, columns, ...rest }) => (
+    <ul
+        className={classNames(
+            'ffe-check-list',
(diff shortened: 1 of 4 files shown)


antidecaf (approved):
Nice! CSS columns would be the logical choice given what it's designed to do, but it still behaves weirdly in general despite having been "supported" by all major browsers for a while already. I'd go the flex route in this case too.

PR #485, opened 2018-10-31
Add missing 'ffe-' prefix to example in ffe-icons

chlunde:
Minor fix in the documentation, add missing `ffe-` prefix to the example so copy & paste works when getting started with ffe :smile:

### DIFF
--- a/packages/ffe-icons/README.md
+++ b/packages/ffe-icons/README.md
@@ -79,7 +79,7 @@ royal blue as the primary color, add the following to your global icon class:
 ```
 .icon {
     fill: "#002776"; // Or, even better:
-    fill: @blue-royal; // assuming use and import of ffe-core variables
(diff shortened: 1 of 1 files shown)


antidecaf (changes requested):
Thanks for the fix!

One slight change request: The commit message needs to follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.2/#specification) format (see [CONTRIBUTING.md](https://github.com/SpareBank1/designsystem/blob/develop/CONTRIBUTING.md#commit-meldinger)). In this case it should be prefixed with `fix(ffe-icons): `. This is necessary for a new version to be published to npm, as well as for Lerna to generate the changelog. :)

antidecaf (dismissed):
... Come to think of it though, publishing a new version for this isn't technically necessary, so feel free disregard my change request if you wish! :flushed:

kwltrs (approved):

PR #505, opened 2018-11-12
Feature/some kind of card

### LINKED ISSUE #260: New type of card(?) for overview and landing pages

henrikhermansen:
I wanted to reopen #262 , but because I replaced my previous commits with new ones, I wasn't allowed to do that.

I've taken comments, opinions and issues raised in the previous PR into consideration when I reworked the styling and components for the new cards. With this, I hope we can come a lot closer to being able to merge this.

I've also taken the liberty of naming them. While these are mere suggestions, I do feel they make sense, as they reflect what's significant about each card.
Speaking of names, I have controversially named the base class and component Card Base. My thought behind this is that it's not really a card - it's a base. This also distinguishes it more clearly from the actual card components. Thoughts?

Closes #260

### DIFF
--- a/packages/ffe-cards-react/src/CardBase.spec.js
+++ b/packages/ffe-cards-react/src/CardBase.spec.js
@@ -1,0 +1,33 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import CardBase from './CardBase';
+
+const getWrapper = props => shallow(<CardBase {...props} />);
+const children = <div>Hello world</div>;
+
+describe('CardBase', () => {
+    it('should render a div element with correct class and children inside', () => {
+        const wrapper = getWrapper({ children });
+
+        expect(wrapper.find('div').exists()).toBe(true);
+        expect(wrapper.hasClass('ffe-card-base')).toBe(true);
+        expect(
+            wrapper
+                .find('.ffe-card-base')
+                .children()
+                .first()
+                .getElement(),
+        ).toEqual(children);
+    });
+
+    it('should render my custom element and custom class', () => {
+        const wrapper = getWrapper({
+            className: 'my-custom-class',
+            element: 'article',
+        });
+
+        expect(wrapper.find('article').exists()).toBe(true);
+        expect(wrapper.hasClass('my-custom-class')).toBe(true);
+    });
+});
--- a/packages/ffe-cards-react/src/IconCard/IconCard.md
+++ b/packages/ffe-cards-react/src/IconCard/IconCard.md
@@ -1,0 +1,47 @@
+Ikonkort er kort med en gitt maksbredde og et venstrestilt ikon.
+Den tilbyr alle underkomponenter (CardName, Title, Subtext og Text) og plasserer dem i
+gitt rekkefølge nedover i kortet, med teksten venstrestilt, til høyre for ikonet.
+
+```jsx
+const { GrafOppIkon } = require('../../../ffe-icons-react');
+
+<IconCard icon={<GrafOppIkon />}>
+    {({ CardName, Title, Subtext, Text }) => (
+        <React.Fragment>
+            <CardName>Kortnavn</CardName>
+            <Title>Tittel</Title>
+            <Subtext>En liten undertekst</Subtext>
+            <Text>Her kan man ha tekst</Text>
+        </React.Fragment>
+    )}
+</IconCard>;
+```
+
+Kortet har to modifiers: condensed og greyCharcoal
+
+```jsx
+const { SparegrisIkon, KryssSirkelIkon } = require('../../../ffe-icons-react');
+
+<React.Fragment>
+    <IconCard icon={<SparegrisIkon />} condensed={true}>
+        {({ Title, Subtext }) => (
+            <React.Fragment>
+                <Title>Sparekonto voksen 25</Title>
+                <Subtext>7 004,00</Subtext>
+            </React.Fragment>
+        )}
+    </IconCard>
+    <IconCard
+        icon={<KryssSirkelIkon style={{ transform: 'rotate(45deg)' }} />}
+        greyCharcoal={true}
+        condensed={true}
+    >
+        {({ Title, Subtext }) => (
+            <React.Fragment>
+                <Title>Start ny sparing</Title>
+                <Subtext>Det er mange måter å spare på</Subtext>
+            </React.Fragment>
+        )}
+    </IconCard>
+</React.Fragment>;
+```
--- a/packages/ffe-cards-react/src/ImageCard/ImageCard.js
+++ b/packages/ffe-cards-react/src/ImageCard/ImageCard.js
@@ -1,0 +1,34 @@
+import React from 'react';
+import classNames from 'classnames';
+import { node, string, func, oneOfType } from 'prop-types';
+
+import CardBase from '../CardBase';
+import * as components from '../components';
+
+const ImageCard = props => {
+    const { className, image, children, ...rest } = props;
+
+    return (
+        <CardBase className={classNames('ffe-image-card', className)} {...rest}>
+            <div className="ffe-image-card__image">
+                <div className="ffe-image-card__image__overlay" />
+                {image}
+            </div>
+            <div className="ffe-image-card__body">
+                {typeof children === 'function'
+                    ? children(components)
+                    : children}
+            </div>
+        </CardBase>
+    );
+};
+
+ImageCard.propTypes = {
+    className: string,
+    /** A rendered image */
+    image: node.isRequired,
+    /** Function that's passed available sub-components as arguments, or regular children */
+    children: oneOfType([func, node]),
+};
+
+export default ImageCard;
--- a/packages/ffe-cards-react/src/ImageCard/ImageCard.md
+++ b/packages/ffe-cards-react/src/ImageCard/ImageCard.md
@@ -1,0 +1,43 @@
+Bildekort er kort med en gitt maksbredde, minimumshøyde, og et topplasert bilde.
+Den tilbyr alle underkomponenter (CardName, Title, Subtext og Text) og plasserer dem i
+gitt rekkefølge nedover i kortet, med teksten midtstilt, nedenfor bildet.
+
+```jsx
+const { GrafOppIkon } = require('../../../ffe-icons-react');
+
+<React.Fragment>
+    <ImageCard
+        image={
+            <img
+                src="https://www.sparebank1.no/content/dam/SB1/foto/profilbilder-liggende/ung-i-sofa.jpg.thumb.1280.1280.jpg"
+                alt=""
+            />
+        }
+    >
+        {({ CardName, Title, Subtext, Text }) => (
+            <React.Fragment>
+                <CardName>Kortnavn</CardName>
+                <Title>Tittel</Title>
+                <Subtext>En liten undertekst</Subtext>
+                <Text>
+                    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed
+                    do eiusmod tempor incididunt ut labore et dolore magna
+                    aliqua. Ut enim ad minim veniam, quis nostrud exercitation
+                    ullamco laboris nisi ut aliquip ex ea commodo consequat.
+                </Text>
+            </React.Fragment>
+        )}
+    </ImageCard>
+    <ImageCard
+        image={
+            <img
+                src="https://www.sparebank1.no/content/dam/SB1/foto/profilbilder-liggende/ung-i-sofa.jpg.thumb.1280.1280.jpg"
+                alt=""
+                style={{ position: 'relative', bottom: '50px' }}
+            />
+        }
+    >
+        {({ Title }) => <Title>Tittel</Title>}
+    </ImageCard>
+</React.Fragment>;
+```
--- a/packages/ffe-cards-react/src/components/Subtext.js
+++ b/packages/ffe-cards-react/src/components/Subtext.js
@@ -1,0 +1,16 @@
+import React from 'react';
+import classNames from 'classnames';
+import { string } from 'prop-types';
+
+import ComponentBase from './ComponentBase';
+
+const Subtext = ({ className, ...rest }) => (
+    <ComponentBase
+        className={classNames('ffe-card-component--subtext', className)}
+        {...rest}
+    />
+);
+
+Subtext.propTypes = { className: string };
+
+export default Subtext;
--- a/packages/ffe-all.less
+++ b/packages/ffe-all.less
@@ -5,6 +5,7 @@
 @import './ffe-buttons/less/buttons';
 @import './ffe-cards/less/ffe-link-card';
 @import './ffe-cards/less/ffe-product-card';
+@import './ffe-cards/less/cards';
 @import './ffe-chart-donut-react/src/ffe-chart-donut';
 @import './ffe-context-message/less/ffe-context-message';
 @import './ffe-datepicker/less/datepicker';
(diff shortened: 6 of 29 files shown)


selbekk:
Can we get a screenshot of how this will look? :)

henrikhermansen:
image
image
image
image
image
image
image

packages/ffe-cards-react/src/IconCard/IconCard.md
```diff
@@ -0,0 +1,49 @@
+Ikonkort er kort med en gitt maksbredde og et venstrestilt ikon.
+Den tilbyr alle underkomponenter (CardName, Title, Subtext og Text) og plasserer dem i
+gitt rekkefølge nedover i kortet, med teksten venstrestilt, til høyere for ikonet.
```
ivarni: *høyre

packages/ffe-cards-react/src/IconCard/IconCard.md
```diff
@@ -0,0 +1,49 @@
...
+
+```jsx
+const { GrafOppIkon } = require('../../../ffe-icons-react');
+
+<React.Fragment>
```
ivarni: Tror ikke du trenger å wrappe `IconCard` i `Fragment`?

packages/ffe-cards-react/src/ImageCard/ImageCard.md
```diff
@@ -0,0 +1,43 @@
...
+    <ImageCard
+        image={
+            <img
+                src="https://www.sparebank1.no/content/dam/SB1/foto/profilbilder-liggende/ung-i-sofa.jpg.thumb.1280.1280.jpg"
+                style={{ position: 'relative', bottom: '50px' }}
+            />
+        }
+    >
+        {({ Title }) => (
+            <React.Fragment>
```
ivarni: Her trenger du heller ikke `Fragment`

ivarni (dismissed):

packages/ffe-cards-react/src/IconCard/IconCard.md
```diff
@@ -0,0 +1,49 @@
...
+            <React.Fragment>
+                <CardName>Kortnavn</CardName>
+                <Title>Tittel</Title>
+                <Subtext>En liten undertekst</Subtext>
+                <Text>Her kan man ha tekst</Text>
+            </React.Fragment>
+        )}
+    </IconCard>
+</React.Fragment>;
+```
```
selbekk, line 18: I guess you could drop Fragment around `<IconCard />`

packages/ffe-cards-react/src/ImageCard/ImageCard.js
```diff
@@ -0,0 +1,33 @@
...
+            </div>
+        </CardBase>
+    );
+};
+
+ImageCard.propTypes = {
+    className: string,
+    /** A rendered image */
+    image: node.isRequired,
+    children: oneOfType([func, node]),
```
selbekk, line 31:
```suggestion
    /** Function that's passed available sub-components as arguments, or regular children */
    children: oneOfType([func, node]),
```

packages/ffe-cards-react/src/ImageCard/ImageCard.md
```diff
@@ -0,0 +1,43 @@
...
+Den tilbyr alle underkomponenter (CardName, Title, Subtext og Text) og plasserer dem i
+gitt rekkefølge nedover i kortet, med teksten midtstilt, nedenfor bildet.
+
+```jsx
+const { GrafOppIkon } = require('../../../ffe-icons-react');
+
+<React.Fragment>
+    <ImageCard
+        image={
+            <img src="https://www.sparebank1.no/content/dam/SB1/foto/profilbilder-liggende/ung-i-sofa.jpg.thumb.1280.1280.jpg" />
```
selbekk:
```suggestion
            <img src="https://www.sparebank1.no/content/dam/SB1/foto/profilbilder-liggende/ung-i-sofa.jpg.thumb.1280.1280.jpg" alt="" />
```

packages/ffe-cards-react/src/ImageCard/ImageCard.md
```diff
@@ -0,0 +1,43 @@
...
+                    ullamco laboris nisi ut aliquip ex ea commodo consequat.
+                </Text>
+            </React.Fragment>
+        )}
+    </ImageCard>
+    <ImageCard
+        image={
+            <img
+                src="https://www.sparebank1.no/content/dam/SB1/foto/profilbilder-liggende/ung-i-sofa.jpg.thumb.1280.1280.jpg"
+                style={{ position: 'relative', bottom: '50px' }}
```
selbekk, line 36: What's this for?

packages/ffe-cards-react/src/ImageCard/ImageCard.md
```diff
@@ -0,0 +1,43 @@
...
+                    aliqua. Ut enim ad minim veniam, quis nostrud exercitation
+                    ullamco laboris nisi ut aliquip ex ea commodo consequat.
+                </Text>
+            </React.Fragment>
+        )}
+    </ImageCard>
+    <ImageCard
+        image={
+            <img
+                src="https://www.sparebank1.no/content/dam/SB1/foto/profilbilder-liggende/ung-i-sofa.jpg.thumb.1280.1280.jpg"
```
selbekk, line 34:
```suggestion
                src="https://www.sparebank1.no/content/dam/SB1/foto/profilbilder-liggende/ung-i-sofa.jpg.thumb.1280.1280.jpg"
                alt=""
```

packages/ffe-cards-react/src/IconCard/IconCard.md
henrikhermansen: Sant, fixed

packages/ffe-cards-react/src/IconCard/IconCard.md
henrikhermansen: Nei, riktig. Jeg hadde to på et tidspunkt, og glemte å fjerne fragment da jeg fjernet det andre kortet. Fixet nå.

packages/ffe-cards-react/src/ImageCard/ImageCard.md
henrikhermansen: Good catch! Jeg hadde flere subcomponents på et tidspunkt. Glemte å fjerne Fragment da jeg gikk ned til bare det ene. Fixed.

packages/ffe-cards-react/src/components/Subtext.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import classNames from 'classnames';
+import { string } from 'prop-types';
+
+import ComponentBase from './ComponentBase';
+
+const Subtext = ({ className, ...rest }) => (
```
selbekk, line 7: Is `<SubTitle />` a better name? Subtext is, according to [Wikipedia](https://en.wikipedia.org/wiki/Subtext):

> Subtext or undertone is any content of a creative work which is not announced explicitly by the characters or author, but is implicit or becomes something understood by the observer of the work as the production unfolds.

selbekk (dismissed):
A few typos and minor stuff, but I like the approach!

I think the pattern of providing sub-components as props was pretty interesting. It's a new approach, and I like it - but does it bring any benefits over just exporting them as a part of the main API?

packages/ffe-cards-react/src/IconCard/IconCard.md
selbekk: Go through all the examples to ensure they're all wrapper-free if they don't require it. Makes for better reading when you look at the example code :)

henrikhermansen:
> does it bring any benefits over just exporting them as a part of the main API?

Good question. While I currently export all sub components through the render prop, I imagine we might make some cards which only allow certain sub components. This way makes it easy to perform such limitations.

On the other hand. If this is never going to be the case, I'm not sure which approach is better. One might argue the API gets a little obfuscated like this, as you won't get any help from the IDE as to which components you have available.

ivarni:
I like that it's harder to use the sub components outside of the context they're meant to be used in (cards). The more limited an exposed API can be the better in my opinion, and even if some cards might not have any restrictions on subcomponents this makes it less tempting to use them outside of and actual card-implementation.

packages/ffe-cards-react/src/ImageCard/ImageCard.md
henrikhermansen: Repositioning the image within the boundary, as the parent has restricted width and height.
It's just to show how easy it is to move it around :)

packages/ffe-cards-react/src/components/Subtext.js
henrikhermansen: Good question. I took all the sub component names from @adidick's https://github.com/SpareBank1/designsystem/pull/262#issuecomment-392735553

selbekk:
A different issue - P-tags are not allowed inside BUTTON tags, according to the internet: https://stackoverflow.com/questions/26002381/p-tag-in-button-tag-not-allowed

If the card is a button, we would need to change the default element of each of the sub-components -they need to be spans or something similar.

packages/ffe-cards-react/src/IconCard/IconCard.md
henrikhermansen: Yeah, I saw there was one more in TextCard.md. Fixed locally.

packages/ffe-cards-react/src/ImageCard/ImageCard.js
henrikhermansen: Thanks, fixed in all of the three cards

packages/ffe-cards-react/src/ImageCard/ImageCard.md
henrikhermansen: Fixed

packages/ffe-cards-react/src/ImageCard/ImageCard.md
henrikhermansen: Fixed

henrikhermansen:
> A different issue - P-tags are not allowed inside BUTTON tags, according to the internet: https://stackoverflow.com/questions/26002381/p-tag-in-button-tag-not-allowed
> 
> If the card is a button, we would need to change the default element of each of the sub-components -they need to be spans or something similar.

Good point. I changed them to be spans as default with display block. Looks exactly the same. I don't know if there are any semantic problems to that approach, but I hope not.

selbekk:
I think we should be good. Of course, the client could screw this up themselves, but I think React should issue a warning. 

Speaking of - are there any relevant aria-tags we should add to these?

ivarni:
Isn't a `<span>` with display block basically a `<div>`?

packages/ffe-cards-react/src/components/Subtext.js
adidick-zz: These names are totally up for improvement so I'm open to any suggestions that make more sense from your perspective.

selbekk:
According to the spec, buttons can only have what's called "phrasing content" inside of it (read more [here](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content)). In other words, to follow the spec, you can't use a `<div />`, because it falls in a different category. It'll work, for sure, but the "correct" way to do it is to use spans.

packages/ffe-cards-react/src/components/Subtext.js
henrikhermansen: To help reduce misunderstandings, I think it's an advantage if ux and developers use the same component names as much as possible. Whether it be [Subtitle](https://en.wikipedia.org/wiki/Subtitle_(titling)), Subtext or something else.

Any other opinions on which name we should favor?

ivarni: We call them "lead paragraphs" in the regular typography but I suppose that kind of depends on how much text we're going to be putting into them.

antidecaf (commented):
Is a card really a button? This doesn't make sense to me. Buttons only allow "phrasing content" as pointed out above, which means it's basically allowed to contain one sentence of text - not subtitles, paragraphs, etc. Sure we can technically get around the spec using spans in stead of divs and what not, but at that point it feels like the markup is designed to circumvent the semantic representation of the contents.

ivarni:
I think in every single usecase I've had for cards so far links have been the obvious choice over buttons since they're anchors for for other pages (detail-pages, buy-our-product-pages etc) so at least from my standpoint it's not really an issues since I can provide an `element` (which 95% of the time for me would be a react-router `Link`).

The example images above also seem like navigation components to me. In what situations are we going to use these as buttons?

selbekk:
That's probably a question for @adidick - what's the intended usage of these outside of linking to other content?

henrikhermansen:
Good points are being made here! To be honest I can't remember why I defaulted these to buttons. I have the same experience as @ivarni , the cards are pretty much always used with Link or anchor (I can't think of a single example where we used something else).

While they will often be used with element=Link, I wonder what should be the default element? I could put an anchor there, but that calls for a default href too, I suppose. Is it ok to set div as default? Perhaps not have any default element, but require one from the consumer?

ivarni:
I think `<div>` is fine, I'm a fan of working defaults whereever possible.

henrikhermansen:
I just pushed the recently discussed changes: components are now paragraphs, without display styling, and cards are now divs.

This looks almost the same, but the CardBase now has left-aligned text (which is default for divs, while buttons defaults it to center).
TextCard and ImageCard has explicit styling for centering the text.

adidick-zz:
> That's probably a question for @adidick - what's the intended usage of these outside of linking to other content?

That is the only intended use.

Description that lives in the sketch template says:
"Kort er en stor klikkflate, altså en inngang til mer informasjon. Kort har en skygge som alltid skal vises, og blir mer fremtredende på hover og focus state."

ivarni (dismissed):

packages/ffe-cards-react/src/CardBase.spec.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import CardBase from './CardBase';
+
+const getWrapper = props => shallow(<CardBase {...props} />);
+const children = <div>Hello world</div>;
+
+describe('CardBase', () => {
+    it('should render a button element with correct class and children inside', () => {
```
selbekk: No longer renders a button - could you please update the text as well?

selbekk (dismissed):
Things are looking good!

Could you also please add a note on the deprecated cards that they are, in fact, deprecated?

antidecaf (dismissed):

packages/ffe-cards-react/src/CardBase.spec.js
henrikhermansen: Good catch! For sure, will do that when I squash commits :)
(I'll do the deprecation note for the old cards as well)

henrikhermansen:
Deprecation note added, fixed test description text and squashed commits. All ready to be merged :)

selbekk (approved):

PR #508, opened 2018-11-14
Make icons esm enabled

selbekk:
This pull request makes the ffe-icons-react package code-splittable, by providing an esm target as well as a CJS one.

### DIFF
--- a/packages/ffe-icons-react/.babelrc
+++ b/packages/ffe-icons-react/.babelrc
@@ -1,0 +1,23 @@
+{
+    "env": {
+        "development": {
+            "presets": [
+                ["env", {
+                    "targets": { "browsers": "ie 11, last 2 versions" }
+                }],
+                "stage-3",
+                "react"
+            ]
+        },
+        "es": {
+            "presets": [
+                ["env", {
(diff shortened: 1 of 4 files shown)


packages/ffe-icons-react/.babelrc
```diff
@@ -0,0 +1,23 @@
...
+                ["env", {
+                    "modules": false,
+                    "targets": { "browsers": "ie 11, last 2 versions" }
+                }],
+                "stage-3",
+                "react"
+            ]
+        }
+    }
+}
```
selbekk, line 23: This can be simplified greatly by upgrading to Babel 7. I'll attack that task tomorrow.

henrikhermansen: Yes! Do it! :)
Beware you might need a dependency `"babel-core": "7.0.0-bridge.0"`, unless jest has released a new version that doesn't depend on babel 6.

ivarni (approved):
noise 1

selbekk:
This could (should?) probably be done for all of our components, in order to improve the final bundle size even more.

PR #524, opened 2018-11-23
fix(ffe-radio-button-react): show error message correctly for RadioSw…

joannae:
…itch.

If selectedValue is not undefined, the red circle for validation error should only be around the button that is selected. If selectedValue is undefined, the red circle is around both buttons. The error text below the buttons is shown the same way as before.

### DIFF
--- a/packages/ffe-radio-button-react/src/RadioSwitch.js
+++ b/packages/ffe-radio-button-react/src/RadioSwitch.js
@@ -1,5 +1,5 @@
 import React, { Fragment } from 'react';
-import { bool, oneOfType, string } from 'prop-types';
+import { bool, oneOf, oneOfType, string } from 'prop-types';
 import classNames from 'classnames';
 
 import BaseRadioButton from './BaseRadioButton';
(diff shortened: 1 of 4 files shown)


packages/ffe-radio-button-react/src/RadioSwitch.js
```diff
@@ -13,9 +13,12 @@ const RadioSwitch = props => {
         rightValue,
         condensed,
         dark,
+        'aria-invalid': ariaInvalid,
         ...rest
     } = props;
 
+    const noneSelected = props.selectedValue === undefined;
```
ivarni: Since we're dealing with javascript, it would be great if every falsy value passed this check. 
So basically, consider `null` and `''` for `selectedValue` to also mean that nothing is selected. Be vary that `0` is something that should be considered a valid selected value though.

Maybe
```javascript
const noneSelected = [null, undefined, ''].includes(props.selectedValue);
```

I think that would reduce the potential for suprises for consumers.

ivarni:
The [travis CI build](https://travis-ci.org/SpareBank1/designsystem/builds/458834522) seems to be breaking

selbekk (dismissed):
A missing empty line - otherwise, this looks good :)

packages/ffe-radio-button-react/src/RadioSwitch.js
joannae: I I completely agree but if wasn't clear to me from the code that this is the wanted behaviour so I have now added tests that describe this.

ivarni (approved):

selbekk:
:tada:

PR #527, opened 2018-11-27
Add toggle switch component

### LINKED ISSUE #519: Switch/toggle
Vi jobber med å få ut Push notifikasjoner i mobilbanken, og har behov for en komponent som har lik design og oppførsel som en native switch/toggle. 

I første skisse så har vi brukt en checkmark men det blir litt misvisende for konteksten vår. 

checkbox

Forslag til ny komponent i bruk:

push-switch

antidecaf:
This PR adds a new toggle switch component. It looks like this:

Default:
image

Checked/selected:
image

**It's not yet decided whether or not this will be merged.** Some questions have arisen along the way and need to be answered first:

- Does the toggle switch pattern make sense in a desktop UI? This component renders a switch on any screen size. Should the switch in stead be replaced by a different pattern on large screens, and if so, which one?

- Guidelines are not yet written, but should be if this is to be merged. When to use a toggle switch vs radio buttons/checkboxes should be very clear. According to [NN Group guidelines](https://www.nngroup.com/articles/toggle-switch-guidelines/) the toggle switch pattern demands immediate results, i.e. they should not be used in forms that require users to click a button for changes to take effect. Any input on this? Anything else?

- How to handle the focus state? This implementation includes the regular focus outline (same one used in form elements), but it feels a bit off given that switches generally don't have an outline in a native mobile context. Do we:

1. Keep it as is?
2. Show the outline only on large screens?
3. Remove the outline altogether? (If so, what consequenses does it have for accessibility?)

- How about the hover state? I've added the focus style on hover for now.

Hover/focus state:
image
image

Any input is welcome!

Fixes #519

### DIFF
--- a/packages/ffe-form-react/src/ToggleSwitch.js
+++ b/packages/ffe-form-react/src/ToggleSwitch.js
@@ -1,0 +1,55 @@
+import React from 'react';
+import { bool, node, string } from 'prop-types';
+import uuid from 'uuid';
+import classNames from 'classnames';
+
+class ToggleSwitch extends React.Component {
+    id = `toggle-${uuid.v4()}`;
+
+    render() {
+        const { children, className, alignRight, id, ...rest } = this.props;
+
+        return (
+            <div
+                className={classNames(
+                    'ffe-toggle-switch',
+                    { 'ffe-toggle-switch--align-right': alignRight },
+                    className,
+                )}
+            >
+                <input
+                    className="ffe-toggle-switch__input"
+                    type="checkbox"
+                    id={id || this.id}
+                    {...rest}
+                />
+                <label
+                    className="ffe-toggle-switch__label"
+                    htmlFor={id || this.id}
+                >
+                    {children}
+                </label>
+            </div>
+        );
+    }
+}
+
+ToggleSwitch.propTypes = {
+    /** The label text */
+    children: node.isRequired,
+    /** Any extra classes */
+    className: string,
+    /** Overrides the generated id with a custom one */
+    id: string,
+    /** Overrides the value attribute of the input with a custom one */
+    value: string,
+    /** Aligns the switch to the right */
+    alignRight: bool,
+};
+
+ToggleSwitch.defaultProps = {
+    alignRight: false,
+    value: 'on',
+};
+
+export default ToggleSwitch;
--- a/packages/ffe-form-react/src/ToggleSwitch.md
+++ b/packages/ffe-form-react/src/ToggleSwitch.md
@@ -1,0 +1,23 @@
+Toggle-switcher er av/på-switcher. De tilbyr brukerne valget mellom to avhengige valg og de har alltid en default verdi. De skal brukes i kontekster hvor systemet gir umiddelbare resultater. De skal ikke brukes dersom du har behov for en lagre-knapp i tillegg. I tillegg skal navnet være kort og konsist.
+
+Toggle-switcher kan erstatte to radioknapper eller én enkelt checkbox når en bruker skal velge mellom to avhengige states.
+
+```js
+<ToggleSwitch>Jeg vil gjerne ha reklame</ToggleSwitch>
+```
+
+Switchen kan skrus på som utgangspunkt med `defaultChecked`:
+
+```js
+<ToggleSwitch defaultChecked={true}>
+    Send meg interessant informasjon
+</ToggleSwitch>
+```
+
+Switch alignet til høyre for labelteksten:
+
+```js
+<ToggleSwitch alignRight={true}>
+    Abonner på ulidelig spennende oppdateringer
+</ToggleSwitch>
+```
--- a/packages/ffe-form-react/src/ToggleSwitch.spec.js
+++ b/packages/ffe-form-react/src/ToggleSwitch.spec.js
@@ -1,0 +1,79 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import ToggleSwitch from './ToggleSwitch';
+
+const getWrapper = props =>
+    shallow(<ToggleSwitch {...props}>Hello world</ToggleSwitch>);
+
+describe('<ToggleSwitch />', () => {
+    it('should render a checkbox', () => {
+        const wrapper = getWrapper();
+        expect(wrapper.find('input[type="checkbox"]').exists()).toBe(true);
+    });
+
+    it('should render a default value if passed', () => {
+        let wrapper = getWrapper();
+
+        expect(wrapper.find('input').prop('checked')).toBe(undefined);
+
+        wrapper = getWrapper({ checked: true });
+
+        expect(wrapper.find('input').prop('checked')).toBe(true);
+    });
+
+    it('should apply the same id to <label> and <input>', () => {
+        const wrapper = getWrapper({ name: 'Some text goes here' });
+
+        expect(wrapper.find('label').prop('htmlFor')).toBe(
+            wrapper.find('input').prop('id'),
+        );
+    });
+
+    it('should apply an unique id for each instance', () => {
+        const wrapper1 = getWrapper({ name: 'Some text goes here' });
+        const wrapper2 = getWrapper({ name: 'Some other text goes here' });
+
+        expect(wrapper1.find('input').prop('id')).not.toBe(
+            wrapper2.find('input').prop('id'),
+        );
+    });
+
+    it('should not change id when re-rendering an instance', () => {
+        const wrapper = getWrapper({ name: 'Some text goes here' });
+        const id = wrapper.find('input').prop('id');
+
+        // ShallowWrapper.update() alone does not force a re-render, it seems.
+        wrapper.instance().forceUpdate();
+        wrapper.update();
+        wrapper.setProps({});
+
+        expect(wrapper.find('input').prop('id')).toBe(id);
+    });
+
+    it('should allow to override the id', () => {
+        const id = 'this-is-not-a-generated-id';
+        const wrapper = getWrapper({ id, name: 'Some text goes here' });
+
+        expect(wrapper.find('label').prop('htmlFor')).toBe(id);
+        expect(wrapper.find('input').prop('id')).toBe(id);
+    });
+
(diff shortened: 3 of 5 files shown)


ivarni:
I don't think focus-states are very relevant on mobile? I'm not even sure how I focus an input control without triggering it on my phone. Or are there accessibility tools out there that rely on focus styling?

packages/ffe-form-react/src/ToggleSwitch.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { bool, node, string } from 'prop-types';
+import { v4 as hash } from 'uuid';
+import classNames from 'classnames';
+
+export default function ToggleSwitch(props) {
+    const { children, className, right, ...rest } = props;
+
+    const id = props.id || `toggle-${hash()}`;
```
selbekk: This will receive a new ID on each render. Instead, use a class component, and set the ID as an instance variable

packages/ffe-form-react/src/ToggleSwitch.js
```diff
@@ -0,0 +1,43 @@
...
+    right: false,
+};
+
+ToggleSwitch.propTypes = {
+    /** The label text */
+    children: node.isRequired,
+    /** Any extra classes */
+    className: string,
+    /** Align the switch to the right */
+    right: bool,
```
selbekk: perhaps `alignRight` is a more descriptive name?

selbekk: Also - `id` is missing from the propTypes.

selbekk: Also also - should `value` also be specified here? It's the main API to the component, could be nice to include in the docs?

packages/ffe-form-react/src/ToggleSwitch.md
```diff
@@ -0,0 +1,19 @@
...
+<ToggleSwitch value="on" defaultChecked={true}>
+    Send meg interessant informasjon
+</ToggleSwitch>
+```
+
+Switch alignet til høyre for labelteksten:
+
+```js
+<ToggleSwitch value="on" right={true}>
+    Abonner på ulidelig spennende oppdateringer
```
selbekk, line 21: :joy:

selbekk (changes requested):
All in all, this looks very good! Should we also add some usage information, like when this is a good choice over the very similar RadioSwitch?

antidecaf:
> I don't think focus-states are very relevant on mobile? I'm not even sure how I focus an input control without triggering it on my phone. Or are there accessibility tools out there that rely on focus styling?

You're right, it's mainly an issue in desktop browsers. But on a second glance it looks less weird than anticipated with the hover state that is identical to the focus state.

packages/ffe-form-react/src/ToggleSwitch.js
antidecaf: Will fix!

packages/ffe-form-react/src/ToggleSwitch.js
antidecaf: Will fix!

antidecaf:
> Should we also add some usage information, like when this is a good choice over the very similar RadioSwitch?

Yes, as mentioned in the original comment, guidelines need to be added before this is merged. I'm not sure what exactly to add however, so I'll await some feedback from UX before doing any of that. :)

kwltrs (dismissed):

hodoea (changes requested):
skjermbilde 2018-12-07 kl 13 55 28


Kan dette fungere som guidelines?

selbekk:
Veldig bra @hodoea ! Jeg savner fortsatt en differensiering mellom når man skal bruke en toggle vs en radio switch, though - kanskje verdt å nevne eksplisitt?

hodoea:
My two cents on these questions:

>It's not yet decided whether or not this will be merged. Some questions have arisen along the way and need to be answered first:

>Does the toggle switch pattern make sense in a desktop UI? This component renders a switch on any screen size. Should the switch in stead be replaced by a different pattern on large screens, and if so, which one?

I am bit unsure here, I dont see any problem with using this for desktop as long as we can convey to people how to best use this in the guidelines. But many of my fellow interaction designers don`t agree. They propose that we look for something else on desktop. These are some  examples from Google forms where you use a toggle-switch to "turn on" more features in your form on desktop. And the other example is where you use it to state that something is required, I dont understand why they use this here though...

However I`ve mostly seen the use of toggle-switches for native functionalities (settings on your phone++)

screen recording 2018-12-06 at 12 59 pm

And another example:
screen recording 2018-12-06 at 01 00 pm

>Guidelines are not yet written, but should be if this is to be merged. When to use a toggle switch vs radio buttons/checkboxes should be very clear. According to NN Group guidelines the toggle switch pattern demands immediate results, i.e. they should not be used in forms that require users to click a button for changes to take effect. Any input on this? Anything else?

>How to handle the focus state? This implementation includes the regular focus outline (same one used in form elements), but it feels a bit off given that switches generally don't have an outline in a native mobile context. Do we:

>Keep it as is?
>Show the outline only on large screens?

I propose we show the outline on large screens only if that is possible

>Remove the outline altogether? (If so, what consequenses does it have for accessibility?)

>How about the hover state? I've added the focus style on hover for now.

On mobile you do not have a hover state. I think perhaps this is one of the challenges if we use this component for both mobile and desktop.

antidecaf:
Yes, the hover/focus states are only present on large screens. Looks fine to me now:

ezgif-1-ea1f06cdb031

I've added the guidelines, but won't merge the PR for now, as it has been decided that this component needs to be tested in an app first.

hodoea:
@selbekk hm.. I have not used this component before, and I read what we have of guidelines for this. I dont see why we should use this over a radio button in our forms. I have raised this in our fora on Slack. One designer mentioned that they have many questions in their forms where they have used radio switches. Because of the components size the questions are "drowning" in the form. 
Maybe we can fix this by adding to the guidelines that if you have more than x questions in a form, consider using radio buttons?

Where is a radio switch a very good use? Maybe anyone knows where the need for this come from.. ?

antidecaf:
The RadioSwitch component differs from ToggleSwitch in that it has no default value, and is used as a part of longer forms with submit buttons. Apart from that I figure it has an entirely different feel. ToggleSwitches are designed to imitate physical switches, which I'm assuming is why the strict on/off usage guidelines make sense.

hodoea:
Yes ofc forgot the default value. I was just a little bit confused because of the second statement in the guidelines: typisk i formen "ja eller nei", **eller "av eller på"**.

hodoea:
Yes ofc, I was just a little bit confused because of the second statement in the guidelines: typisk i formen "ja eller nei", **eller "av eller på"*.

hodoea:
Nvm, Adi shared this on the background for it: 

_Radio switch var laget til å tydeligjøre lett valg på mobil. Det var laget for ca. 4 år siden som en del av NN (nye nettsider) og DNA (Forsikring) prosjekter akkurat i starten av å jobbe med den nye profil refresh. Det var testet i bilforsikring kjøpsløp på mobil og fungerte skikkelig bra pga at det var veldig tydelig valg og veldig oversiktlig, plus at det var veldig mobilvennlig. Det var mange runder ift utforming av knappen og det som ble til slutt var en blanding av en som eksisterte før og ny stilen. Jeg synes det er ikke den fineste komponent i at det er for tungt i mørk blå men det kunne vi justere. Ideen er å bare bruke det til lett valg, som “Ja/Nei”. Så snart som det er mer en 3 valg bør man bruke vanlig radio knapper (i en list nedover). Det har alltid vært et spørsmål om hvis vi trenger å ha en så stor og tydelig variant ift helhetlig visuell hierarkiet (f.eks i dag konkurrere den mørke blå med alle andre action knapper på sidene) så det kunne man godt se på. Grunnen at det er mørkt blå er at ideen var at alle “valgt” states burde være likt på tvers av designet og akkurat da brukte vi mørk blå mye mer i designet vårt. I dagens design er det litt for tungt i sammenheng med alle andre knapper osv og kunne være f.eks en mer subtil lys blå, eller bare bruke vanlig radio knapper men passe på at de har en stor klikkflate._
 

I can just add in the documentation for radio switch that if there are more than 3 choices in your list, use radio buttons instead. This was sort of a sidetrack, sorry 😅

selbekk:
Perhaps the difference can be explained like this?

The toggle switch is for "on / off" choices, typically found in settings, while the radio switch is for `yes / no" choices, typically found in forms with a submit button?

hodoea:
Takk. Har allerede skrevet i første avsnitt at man ikke skal bruke toggle-switch dersom man har bruk for en send/lagre knapp i tillegg. Men jeg la det til igjen i forslaget ditt uansett. @selbekk  


skjermbilde 2018-12-07 kl 12 55 29

Sorry for at teksten var i et bilde i starten, har kopiert inn teksten nå :) @antidecaf 

Bruk av toggle-switch

Toggle-switcher er av/på switcher. De tilbyr brukerne valget mellom to avhengige valg og de har alltid en default verdi. De skal brukes i kontekster hvor systemet gir umiddelbare resultater. Det skal ikke brukes dersom du har behov for en lagre/send knapp i tillegg. Navnet du bruker for å beskrive handlingen skal være kort og konsist. 

Toggle-switcher kan erstatte to radioknapper eller én enkelt checkbox når en bruker skal velge mellom to avhengige states.  

Toggle-switch vs radio switch
Bruk en toggle-switch for «av/på» valg, dette er for eksempel vanlig i bl.a innstillinger. Radio switcher derimot brukes for «Ja/Nei» valg som bl.a finnes i skjemaer hvor du i tillegg har behov for en lagre/send knapp.

selbekk:
som jeg skulle gjort det selv @hodoea :laughing: tusen takk!

ivarni:
@antidecaf Are this one pretty much ready to go once the documentation is added?

antidecaf:
@ivarni Yeah, pretty much. But it doesn't have a green light from UX at the moment, it's still being tested in one app afaik. Not sure if it will be merged or not in the end, but for now we can leave it.

PR #530, opened 2018-11-29
492 merge checkbox radio form

ivarni:

### DIFF
--- a/packages/ffe-form-react/src/BaseRadioButton.js
+++ b/packages/ffe-form-react/src/BaseRadioButton.js
@@ -1,0 +1,92 @@
+import React, { Component, Fragment } from 'react';
+import classNames from 'classnames';
+import { bool, node, oneOf, oneOfType, shape, string } from 'prop-types';
+import uuid from 'uuid';
+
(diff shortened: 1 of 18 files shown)


selbekk (dismissed):
Great job!

Should the packages be removed from [`greenkeeper.json`](https://github.com/SpareBank1/designsystem/blob/develop/greenkeeper.json) and  [`commitlint.config.js`](https://github.com/SpareBank1/designsystem/blob/develop/commitlint.config.js) also?

ivarni:
I suppose removing them from greenkeeper is a good idea, dunno about commitlint untill we remove them completely?

ivarni:
Hot damn, of course `ffe-account-selector-react` is now failing because of the removed modules, it seems I can't really do the merge in the same PR as I remove the old code, which I suppose makes sense.

selbekk (approved):
ah right. Yep, this makes sense.

ivarni:
Ninja-merger med en approval fordi #531 ikke kan fikses ferdig-ferdig før dette er published.

PR #531, opened 2018-11-29
refactor(ffe-account-selector-react): New dependencies

ivarni:
BREAKING CHANGE:
This package relied on `ffe-checkbox-react` which has been merged into `ffe-form-react`.

As a result, you need to update your dependencies to use this version. Replace the dependency on `ffe-checkbox-react` with `ffe-form-react`.

*Make sure you're at least on version* `4.3.0` of `ffe-form-react`.

No further action should be required to update this version.


ivarni:
This is going to fail untill #530 is merged and published

selbekk (dismissed):

antidecaf (approved):

PR #532, opened 2018-12-04
feat(ffe-form-react): Allow node as label

### LINKED ISSUE #522: Allow node as label to RadioButtonInputGroup

`ffe-radio-button-react`

The `RadioButtonInputGroup` only allows `string` for the `label` prop but I would like to be able to provide a `node` so I can apply CSS classes to the label. Since the label is only nested inside a basic `<legend>` tag I don't see any reason for not allowing this.

ivarni:
Allow using either a node or a string for the label of a
`RadioButtonInputGroup`.

Closes #522

### DIFF
--- a/packages/ffe-form-react/src/RadioButtonInputGroup.js
+++ b/packages/ffe-form-react/src/RadioButtonInputGroup.js
@@ -90,7 +90,7 @@ RadioButtonInputGroup.propTypes = {
      * prop but provide your own label you should make sure your solution passes
      * acessibility validation using a tool such as aXe DevTools.
      */
-    label: string,
+    label: oneOfType([node, string]),
     /** The name of the radio button */
     name: string,
     /** Change handler, receives value of selected radio button */


kwltrs (approved):

packages/ffe-form-react/src/RadioButtonInputGroup.js
```diff
@@ -90,7 +90,7 @@ RadioButtonInputGroup.propTypes = {
      * prop but provide your own label you should make sure your solution passes
      * acessibility validation using a tool such as aXe DevTools.
      */
-    label: string,
+    label: oneOfType([node, string]),
```
selbekk, line 93: Note that the `legend` tag only accepts ["phrasing content"](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content) - but I guess the React DOM might warn about misuse if it occurs.

selbekk (approved):
Looks good

PR #545, opened 2018-12-17
fix(ffe-form): remove clickable ghost checkboxes

henrikhermansen:
The ghost checkboxes (inputs with `.ffe-hidden-checkbox`) are clickable, which can create some confusing UI. Let's fix that.

Please let me know if you think there are better ways to fix this.


antidecaf (approved):

selbekk (approved):
Looks fine!

Issue #501, opened 2018-11-09
AccountSelectorMulti doesn't work with screen reader


42tte:
#### Which package is this issue related to?
ffe-account-selector-react

#### Describe your issue (screenshots welcome!)
The accountselector does not work with screen reader (VO/Safari) and needs thorough testing and overhauling. Exemple of bugs:
* No information is given on that there is alternatives neighter are they read on focus.
* Not able to type in input field
* Add space event as selector?
* role=combobox is on the wrong element, should be on input
* 

#### What is the expected behavior? 
It should be usable with screen reader

#### What is the actual behavior?
It is not

#### Steps to reproduce
Use a screen reader on https://design.sparebank1.no/styleguidist/index.html#!/AccountSelectorMulti

ivarni:
Instead of continuing to reinvent the wheel, should we just swap to using [downshift](https://github.com/paypal/downshift)?

42tte:
I second that, if no one else takes it upon them to test that the Downshift examples actually work I will try to do it asap (maybe next week).

First quick test of examples 01-04 shows it also have some issues, but generally it works better than ours and it would be a good thing contributing to it.

PR #568, opened 2019-02-09
fix(ffe-account-selector-react): Optimize performance of suggestion list

### LINKED ISSUE #367: AccountSelector 'pagination scrolling', render window, performance optimalisation
ffe-account-selector-react
When populating the component with massive amount of data we should just render a data window that is visible to the scroll position instead of the whole dataset. A lot of data generates too much eventing and virtual dom diffing and JS garbage collection problems so we should use selective rendering techniques to make the component useful for big datasets.
No lagging, no crashing of browsers on limited memory machines when big datasets populate the component. 
Performance problems on big datasets.
#### Steps to reproduce
Populate the component with thousands of rows.

simjes:
Not sure if you still want this, but fixes #367. It might be better to rewrite for #501 first?

The list will require a height, as `react-window` does not automatically shrink if needed bvaughn/react-window#121.

Doesn't look great with the default height when items doesn't max out the height.
image

### DIFF
--- a/packages/ffe-account-selector-react/less/base-selector.less
+++ b/packages/ffe-account-selector-react/less/base-selector.less
@@ -12,7 +12,7 @@
 
     &__expand-button {
         position: absolute;
-        padding : 6px;
+        padding: 6px;
         top: 50%;
         transform: translateY(-50%);
         right: 10px;
@@ -59,15 +59,5 @@
         border: @border;
         border-radius: 2px;
         background: @ffe-white;
-        &-list {
-            left: 0;
-            width: 100%;
-            top: 45px;
-            box-sizing: border-box;
-            z-index: 100;
-            margin: 0;
-            padding: 0;
-            list-style: none;
-        }
     }
 }
--- a/packages/ffe-account-selector-react/package.json
+++ b/packages/ffe-account-selector-react/package.json
@@ -38,7 +38,8 @@
     "classnames": "^2.2.5",
     "prop-types": "^15.6.0",
     "react-auto-bind": "^0.4.2",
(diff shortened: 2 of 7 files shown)


simjes:
I could implement auto sizing if you want me to continue on this PR, f.ex with https://github.com/souporserious/react-measure

henrikhermansen:
This seems to me like a good start. I see the need for improved performance here, and react-window is a good option for that. I do, however, feel that this will only be a good solution if you can implement auto sizing, as it doesn't look nice for just a few accounts (as your screen shot illustrates).

Great if you can keep up the work on this, and I'll review it again when you have implemented auto sizing :)

henrikhermansen (changes requested):
This seems to me like a good start. I see the need for improved performance here, and react-window is a good option for that. I do, however, feel that this will only be a good solution if you can implement auto sizing, as it doesn't look nice for just a few accounts (as your screen shot illustrates).

Great if you can keep up the work on this, and I'll review it again when you have implemented auto sizing :)

simjes:
Abandoning this as it is not up to standards, will work on a new PR instead

ksawerysb1:
@simjes @henrikhermansen har vi noe nytt på denne fronten? Eller en ny komponent som kan søke og scrolle 1000+ resultater?

ksawerysb1:
Liten oppdatering, jobbet litt med å få denne saken rebased på nyeste release, også fikset en del rundt visning og markup:
https://github.com/ksawerysb1/designsystem/commit/a2d1ac1bab266e662db0f598c379d7ab67c6c986

PR #578, opened 2019-02-28
372 deprecated lifecycle callbacks

ivarni:
It's possible that the datepicker could have been changed to not cause an extra render with `componentDidUpdate` but it was the easiest and shortest way to remove the deprecated callbacks without touching too much of the other code.


ivarni:
Note that I did not fix `ffe-account-selector-react`, it went into a loop when I tried and I don't really want to rewrite the entire thing so I'll leave that up to someone else.

antidecaf (approved):

Issue #595, opened 2019-03-26
ffe-tabels-react: Row may be expanded as default


martinhbarnes:
#### Which package is this issue related to?
ffe-tables-react

#### Describe your issue (screenshots welcome!)
TableRowExpanded should have the option of being expanded as default. 
When the expanded content of a TableRowExpandable contains a link to another page, and that other page contains a back-link, users want the expanded row to still be expanded when clicking the back button, and they want the page to be scrolled to that expanded row.

#### What is the expected behavior? 
The row that was expanded is still expanded and the page is scrolled to that expanded row.

#### What is the actual behavior?
The row is not expanded and the scroll position is at the top of the page.

PR #596, opened 2019-03-26
feat(ffe-tables-react): Expandable table rows may be expanded as default

martinhbarnes:
This version adds support for expanding a table row by default. The cells prop in TableRowExpandable may have a value defaultExpanded, that row is then expanded as default and scrolled to on mount.

This functionality is needed eg. when the expanded content is containing a link to another page, and that page has a back button. We then want to keep the position in the table, and have the option still expanded, when pressing the back button.

Fixes: #595

### DIFF
--- a/packages/ffe-tables-react/src/Table.md
+++ b/packages/ffe-tables-react/src/Table.md
@@ -129,6 +129,65 @@ const onSort = ({ sortBy, descending, tableData }) => {};
 />;
 ```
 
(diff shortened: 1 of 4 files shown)


martinhbarnes:
Etter en prat med @ksawerysb1 flytter jeg scrollIntoView ut som en egen prop slik at det skal være mulig å la være rader være expanded by default, og sender inn begge som en prop til TableRowExpandable

ksawerysb1 (approved):

antidecaf (approved):

PR #606, opened 2019-04-12
feat(ffe-decorators-react): Add `alwaysEaseFrom` option to `easePrope…

### LINKED ISSUE #605: Add option for easing from zero

`ffe-decorators-react`

We have a usecase where we're animating a donut-chart using `easeProperties` but in order to better communicate how the value changes we want every property change to ease from 0.

Example:
An eased property starts out as 100, when it's changed to 90 we want it to ease up from 0 to 90 instead of easing down from 100 to 90.

Suggestion:
Add a property to the configuration object, `alwaysEaseFrom`, that sets a fixed value to ease any property from.

```javascript
easeProperties({
    percentage: { initialValue: 100, duration: 0.75, alwaysEaseFrom: 0 },
})(Component)
```

ivarni:
…rties`

Adds a configuration option, `alwaysEaseFrom`, that if set will cause all
updates on the wrapped component to start easing from that value instead
of the previous value.

Closes #605

### DIFF
--- a/packages/ffe-decorators-react/src/ease-properties/easeProperties.js
+++ b/packages/ffe-decorators-react/src/ease-properties/easeProperties.js
@@ -19,7 +19,8 @@ export default (properties = {}) => TargetComponent =>
     // {
     //      somePropertyToEase: {
     //          duration: 0.5, // duration of easing in seconds
-    //          initialValue: 0 // the initial value to start the easing from when the component mounts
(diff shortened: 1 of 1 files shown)


cecilieboe (approved):

ivarni:
I'll just copy this thing over to our codebase and modify it there instead

Issue #614, opened 2019-04-26
ffe-tab-button--selected shoud include aria-pressed=true


magnusjaeger:
ffe-tab-button--selected shoud include aria-pressed=true, to ensure current state is communicated.

exoen:
This issue should be closed as PR #617 is merged

PR #617, opened 2019-05-01
feat(ffe-tabs-react): Add aria-selected to tabs

exoen:
Fixes issue #614

### DIFF
--- a/packages/ffe-tabs-react/src/Tab.js
+++ b/packages/ffe-tabs-react/src/Tab.js
@@ -6,6 +6,8 @@ export default function Tab(props) {
     const { className, selected, condensed, ...rest } = props;
     return (
         <button
+            role="tab"
+            aria-selected={selected}
             className={classNames(
                 'ffe-tab',
                 { 'ffe-tab--selected': selected },
--- a/packages/ffe-tabs-react/src/Tab.spec.js
+++ b/packages/ffe-tabs-react/src/Tab.spec.js
@@ -19,6 +19,16 @@ describe('Tab', () => {
         expect(wrapper.hasClass('ffe-tab--selected')).toBe(true);
     });
 
+    it('should have aria-selected to be true when tab is selected', () => {
+        const wrapper = shallow(<Tab selected={true}>En tab</Tab>);
+        expect(wrapper.prop('aria-selected')).toBe(true);
+    });
+
+    it('should have aria-selected to be false when tab is not selected', () => {
+        const wrapper = shallow(<Tab selected={false}>En tab</Tab>);
+        expect(wrapper.prop('aria-selected')).toBe(false);
+    });
+
     it('passes any prop on to a tab', () => {
         const tab = shallow(<Tab data-analytics-track="logMe">En tab</Tab>);
         expect(tab.prop('data-analytics-track')).toBe('logMe');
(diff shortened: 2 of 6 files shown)


simjes:
Isn't `aria-selected` more appropriate for tabs?

hdaljord:
> Isn't `aria-selected` more appropriate for tabs?

Yes, aria-selected is correct. See https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html

simjes (changes requested):
Change to `aria-selected`

ivarni:
Note that this now fails the build on an a11y lint-rule
```
error  The attribute aria-selected is not supported by the role button. This role is implicit on the element button  jsx-a11y/role-supports-aria-props
```
See [rule description](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/role-supports-aria-props.md)

Looking at the error message and [this example](https://www.stefanjudis.com/blog/aria-selected-and-when-to-use-it/#the-actual-use-case-of-code-aria-selected-code-) it seems to me that you should also add the `role="tab"` attribute before using `aria-selected`. This is consistent with the link @hdaljord provided, where the example is also using that role on the tab-button.

image

So I guess yay for linters.

ivarni:
To follow up on that, we really should require an `aria-controls` prop in the `propType` for the buttons to make sure consumers also provide which tab it controls. There's little use telling a screenreader that this controls a tab without also telling them **which** tab it controls.

exoen:
Yay for linters!

Thanks guys, I'll push an update soon. 

@ivarni ,  should `aria-controls` be created as a seperate issue?

ivarni:
@exoen That's up to you really, it's just a suggestion from me. By making it a required prop we're forcing the developer to at least consider providing it but it's not really a part of the tab component itself since it would be referring to some other element in another part of the DOM. 

If you want to keep the scope down on this PR, I'm sure creating an issue is just fine. Maybe we should have a quick look at the bigger picture here and update the documentation with a full-fledged example of how to use `ffe-tabs-react` buttons with actual tabs instead of the [rather bare-bones example](https://design.sparebank1.no/styleguidist/index.html#!/Fanekomponenter).

Either way, this PR is an improvement once the `tab` role is in place so, `aria-controls` was just a digression really.

exoen:
Then I want to limit this PR to fix #614 as it was intended and rather create new issues for `aria-controls` and documentation with examples. 

Any tips on tools and screen readers to use when developing to actually test and improve the experience when using screen readers?

ivarni:
@exoen I have previously installed and tried to use NVDA which is free but it's not a simple task to learn to use it unfortunately, so I mostly rely on the react-a11y linter, reading about usability and trying to best implement the (sometimes conflicting) advice they give. 

There's also browser plugins like Wave and Axe and Chrome devtools have the Audit tab which can check for Accessibility but to actual test how the experience is for someone with a screen reader is something I find incredible difficult. None of those tools would have picked up one the fact that we were missing `aria-selected` and the `tab` role because there's no way they could differentiate between a regular button and a button being used as a tab control.

TL;DR: It's very difficult, unfortunately.

ivarni (commented):
👍

simjes (approved):

ivarni (approved):

PR #620, opened 2019-05-08
More ts definition files

hdaljord:
This PR adds typescript definition files, according to issue #598, for the following ffe-packages:
* ffe-message-box-react
* ffe-context-message-react
* ffe-lists-react
* ffe-buttons-react
* ffe-core-react
* ffe-dropdown-react
* ffe-form-react
* ffe-tables-react
* ffe-tabs-react

It also adds a fix to ffe-system-message-react ts definitions.

Please note that there is a breaking change in  ffe-form-react, where a few value-parameters have changed propType-definitions as well, according to HTML input value being a DOMString-attribute (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio).

### DIFF
--- a/packages/ffe-buttons-react/README.md
+++ b/packages/ffe-buttons-react/README.md
@@ -10,3 +10,8 @@ npm install --save @sb1/ffe-buttons-react
 
 Run Styleguidist from the repository root to see live examples and documentation,
 or see the markdown files next to the component code in `src/`.
+
+## TypeScript definition files
+
+This component supports TypeScript - please update `index.d.ts` if you change any
+of the external methods or properties in this component.
--- a/packages/ffe-buttons-react/package.json
+++ b/packages/ffe-buttons-react/package.json
@@ -9,19 +9,22 @@
   "author": "SpareBank 1",
   "files": [
     "lib",
-    "es"
+    "es",
+    "types"
   ],
   "main": "lib/index.js",
   "module": "es/index.js",
+  "types": "types/index.d.ts",
   "sideEffects": false,
   "repository": {
     "type": "git",
     "url": "ssh://git@github.com:SpareBank1/designsystem.git"
   },
   "scripts": {
-    "build": "npm run build:cjs && npm run build:es",
+    "build": "npm run build:cjs && npm run build:es && npm run build:types",
     "build:cjs": "babel -d lib/. --root-mode=upward  --env-name=cjs --ignore=*.spec.js src/.",
     "build:es": "babel -d es/. --root-mode=upward --env-name=es --ignore=*.spec.js src/.",
+    "build:types": "copyfiles -f src/index.d.ts types/",
     "lint": "eslint src/.",
     "test": "jest",
     "test:watch": "jest --watch"
--- a/packages/ffe-buttons-react/src/index.d.ts
+++ b/packages/ffe-buttons-react/src/index.d.ts
@@ -1,0 +1,97 @@
+import * as React from 'react';
+
+export interface MinimalBaseButtonProps extends React.HTMLProps<HTMLElement> {
+    className?: string;
+    element?: HTMLElement | string;
+    innerRef?: React.RefObject<HTMLElement>;
+}
+
+export interface BaseButtonProps extends MinimalBaseButtonProps {
+    children?: React.ReactNode;
+    ariaLoadingMessage?: string;
+    condensed?: boolean;
+    disabled?: boolean;
+    isLoading?: boolean;
+    leftIcon?: React.ReactNode;
+    rightIcon?: React.ReactNode;
+    dark?: boolean;
+}
+
+export interface ActionButtonProps extends BaseButtonProps {
+    ghost?: boolean;
+}
+
+export interface BackButtonProps extends MinimalBaseButtonProps {
+    children?: React.ReactNode;
+    dark?: boolean;
+}
+
+export interface ButtonGroupProps {
+    className?: string;
+    thin?: boolean;
+    inline?: boolean;
+}
+
+export interface ExpandButtonProps extends MinimalBaseButtonProps {
+    children: React.ReactNode;
+    closeLabel?: string;
+    leftIcon?: React.ReactNode;
+    rightIcon?: React.ReactNode;
+    isExpanded: boolean;
+    onClick: (e: React.MouseEvent | undefined) => void;
+}
+
+export interface InlineExpandButtonProps {
+    children?: React.ReactNode;
+    innerRef?: React.RefObject<HTMLElement>;
+    isExpanded: boolean;
+    onClick: (e: React.MouseEvent | undefined) => void;
+}
+
+export interface PrimaryButtonProps extends BaseButtonProps {}
+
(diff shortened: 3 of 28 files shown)


packages/ffe-form-react/src/RadioSwitch.js
```diff
@@ -61,13 +61,13 @@ RadioSwitch.propTypes = {
     /** The label of the choice to the left */
     leftLabel: string.isRequired,
     /** The value of the choice to the left */
-    leftValue: oneOfType([bool, string]).isRequired,
+    leftValue: oneOfType([number, string]).isRequired,
```
ivarni: Not a fan of this tbh. for a component called a "switch", using a `bool` for the state/model value really is the only sensible choice, and React will make sure it ends up in the DOM as a DOMString because that is what we're using React for in the first place... abstracting *away* the DOM.

I don't agree that basing `propTypes` in React on what props are supposed to be in the DOM because the mental model React wants us to use is to not be aware there even is a DOM.

packages/ffe-form-react/src/BaseRadioButton.js
```diff
@@ -78,13 +86,13 @@ BaseRadioButton.propTypes = {
     /** The name of the radio button */
     name: string.isRequired,
     /** The selected value of the radio button set */
-    selectedValue: oneOfType([bool, string]),
+    selectedValue: oneOfType([number, string]),
```
ivarni: Same here, for a component that really only has two active states, it makes very little sense not to allow the model to have a boolean value

packages/ffe-form-react/src/RadioSwitch.js
ivarni: I mean, we're already using `onChange` more or less everywhere even though React's `onChange` does not act like the DOM `change` event at all but more like the `input`.

We should not have to worry about the restraints put on us by the DOM.

hdaljord:
@ivarni, I'll try to answer all your comments at once here.

When it comes to the RadioSwitch component, I agree, as this is a wrapper that has two radio buttons, which both cannot have the same value, it should allow boolean.

But, when the RadioSwitch component just creates two BaseRadioButtons it will create two HTML input elements with the same type as comes from RadioSwitch. In BaseRadioButton it does not give any sense to take bool as input as it really is used as a string anyway.

On a general basis, I must add that I, personally, find it easier and more predictable to work with components that I perceive as wrappers for HTML elements that does not use a common HTML attribute name as something else, or something similar, but not the same. I would expect a value for a radio button component to behave in the same way as the value on the input element does. Therefore I think it is a bad habit to abstract us away from the HTML DOM by using the same attribute names.

I'd just like to add some more background for why I want to do this change:
The reason for adding type definitions for typescript is to make it easier to use the common components, and to avoid mistakes. A lot of our react components supports passing all the properties to the generated HTML element as attributes. To support passing these props when using typescript, we have two possible solutions:
* A: Allow all properties (using [propName: string]: any;)
* B: Allow all valid attributes for the resulting HTML-element

I think A is a bad idea, because then we supports everything, even though the browser cannot make any sense of it. B will allow us to pass all properties that makes any sense to the HTML element, thus it helps the developer to write valid code. (We want to write valid html.)

B can be implemented the easy way, by extending the interfaces from React's type definitions (HTMLAttributes with friends), or we can implement them ourselves and make our own definitions. Just to create a wrapper component for a standard HTML element, I think this will be unnecessary amount of code/manual written type definitions, which we would also have to maintain if they change later.

I agree that we should not make any big rewrites of the javascript react components just to add type definitions for typescript, but in this case I think that this is such a small change, and in line with HTML, that it is worth it.

ivarni:
I can see the argument for only allowing `value` types of the same instance that you'd get from reading them back out of the syntethic event provided in e.g. `onChange`, but then we should support *only* `string`.  Just replacing `bool` with `number` makes no sense in that context.

```jsx
<input 
    onChange={(e) =>console.log(e.target.value)}
    value={1}
/>
```
Would be printing values of type string. But if we want to use that argument then we should take it all the way and just go `string` everywhere. I don't agree that it's a good idea, though, because I want to be able to use proper types in my model, which is my state and where the `value` prop I am sending to an input control comes from. 

I already have to translate from `string` when I read changes out of events and it would be cumbersome having to convert them *back* to strings again to use them in controlled inputs. Especially since React already does that for me. 

```jsx
<input
    type="radio"
    value={true}
/>
```
will yield the exact same partial DOM tree as
```jsx
<input
    type="radio"
    value="true"
/>
```

The `BaseRadioButton` is not exposed as a part of [the provided API](https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-form-react/src/index.js), it's an implementation detail so consumers will not be passing anything to it in the first place unless they're digging around in the `lib` folder, something we really don't want them to do. Arguably, it shouldn't even need to have any propTypes in the first place since it's only ever (supposed to be) used internally by the components we're actually exposing as part of the API.

hdaljord:
BaseRadioButton is not exposed directly, but the properties it allows is exposed as the wrapper component passes their props directly to BaseRadioButton, which in turn use them as attributes on the input element.

If we should follow your arguments, then the propType of these props should not have been limited to only string or bool, but whatever you want? In that direction, why not a number?

To be able to extend the InputHTMLAttribute<HTMLInputElement>, the value has to be defined as number or string. I think this is weird, but not so weird that I would like to implement all the types in our type definition.

ivarni:
What I'm trying to say here is that it's *not* passed to the input element, it's passed to a JSX component and then converted to a string as a part of being rendered as an HTML input element. They're not the same. The JSX we're writing is simply syntactic sugar for 
```javascript
React.createElement('input', { value: true }, null)
```
It's not HTML.

We've had a philosophy for most of the codebase to follow the principle of least surprise and have our custom components work as closesly to the native React ones as possible, that's why we introduced that principle of not renaming attributes (like renaming `aria-valid` to `ariaValid` or `valid` like parts of the codebase unfortunately still do). 

If anyone wants to use `0` or `1` for their radio-switch state, then by all means I see no reason to not let them do that. We just didn't anticipate that usecase. By all means add `number` if you have a use-case for it but I don't see any argument for removing `bool` just to impose arbitrary limitations of HTML on top of JSX. We have just put JSX on top of HTML so that we shouldn't have to.

I guess we'll have to agree to disagree. Since we've both used variations of the exact same argument in both our last posts.

If the BaseRadioButton propTypes are the problem I vote that we remove them. They serve no real purpose except for someone working directly with the package itself and are not part of the distributed npm package.

hdaljord:
I don't agree that to deviate from how HTML treats and expects props is to follow the principle of least surprise, but I have changed it back again.

The problem is that all props are passed down to the real HTML element, but instead of implementing 34 attributes manually I found a workaround in the type definitions.

ivarni (dismissed):
I've added a few props to some components in very recently merged PRs, I'll get them added to the ts files once everything lands on the master branch, probably by monday.

hdaljord:
Thanks, @ivarni :)

I've added the new props in ffe-buttons-react, as they are on master now, but the one in ffe-tabs-react has not been added.

ivarni (approved):

PR #621, opened 2019-05-08
Dark back button

antidecaf:
Added a dark variety of `BackButton`.

### DIFF
--- a/packages/ffe-buttons-react/src/BackButton.md
+++ b/packages/ffe-buttons-react/src/BackButton.md
@@ -1,3 +1,19 @@
 ```js
 <BackButton onClick={f => f}>Tilbake</BackButton>
 ```
+
+Variant _dark_ for interne løsninger med mørk bakgrunn.
+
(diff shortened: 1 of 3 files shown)


sconrads (approved):

packages/ffe-buttons-react/src/BackButton.md
```diff
@@ -1,3 +1,19 @@
 ```js
 <BackButton onClick={f => f}>Tilbake</BackButton>
 ```
+
+Variant _dark_ for interne løsninger med mørk bakgrunn.
+
+```js { "props": { "className": "sb1ds-example-dark" } }
```
ivarni, line 7: Is this `{ "props": { "className": "sb1ds-example-dark" } }` stuff some kind of magic way to declare props?

I'm also pretty sure you don't really need `React.Fragment`

ivarni: Never mind, figured out the `props` thing, that's neat.

PR #625, opened 2019-05-09
fix(ffe-buttons-react): a11y improvements to `ExpandButton`

### LINKED ISSUE #619: An expanded ExpandButton has no discernible text

`ffe-buttons-react`

Once an `ExpandButton` is expanded it only contains a SVG icon and no text or aria controls.
[this](https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-buttons-react/src/ExpandButton.js#L29) line in the implementation omits the provided `children` so any attempts from a consumer to provide a readable label will be ignored.

This means that the consumer will need to provide an `aria-label` **only** when the button is expanded which is not intuitive.

We should either

a) Make this very explicit in the documentation
or
b) Add a required `labelWhenOpen` prop (or something similar) and use that value of that for the `aria-label` property once the button is in the expanded state.

ivarni:
* The `children` prop is now required, since it provides the button label in the closed state
* A new prop `closeLabel` is introduced for showing an accessible label in the expanded state, where no visible label is rendered. It defaults to "Lukk" but can be overriden

Fixes #619

### DIFF
--- a/packages/ffe-buttons-react/src/ExpandButton.js
+++ b/packages/ffe-buttons-react/src/ExpandButton.js
@@ -3,10 +3,11 @@ import { bool, func, oneOfType, string, node } from 'prop-types';
 import classNames from 'classnames';
 import KryssIkon from '@sb1/ffe-icons-react/lib/kryss-ikon';
 
-const ExpandButton = (props) => {
+const ExpandButton = props => {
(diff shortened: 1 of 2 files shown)


antidecaf (approved):
LGTM!

PR #627, opened 2019-05-09
Add required ariaControls to Tab and TabButton

### LINKED ISSUE #626: Tab and TabButton should require an aria-controls prop
ffe-tabs-react
The tab and tabButton should require `aria-controls` so that consumers provide information of which element it controls to help the experience when using screen readers.
aria-controls should be required
aria-controls is not required and must be set manually by the consumers

exoen:
Adds a required property called ariaControls which is set as `aria-controls` for the element.

This pull request also introduces examples of how to use the tab and tab button with content related to the tabs.

Tab example:

Tab example

TabButton example:

TabButton example

Fixes #626

### DIFF
--- a/packages/ffe-tabs-react/src/Tab.js
+++ b/packages/ffe-tabs-react/src/Tab.js
@@ -26,4 +26,6 @@ Tab.propTypes = {
     condensed: bool,
     /** Additional css classes */
     className: string,
+    /** Id of the element it controls */
+    'aria-controls': string.isRequired,
 };
--- a/packages/ffe-tabs-react/src/Tab.spec.js
+++ b/packages/ffe-tabs-react/src/Tab.spec.js
@@ -29,6 +29,12 @@ describe('Tab', () => {
         expect(wrapper.prop('aria-selected')).toBe(false);
     });
 
+    it('should have aria-controls attribute when passed to it', () => {
+        const controlId = 'controlId';
+        const wrapper = shallow(<Tab aria-controls={controlId}>En tab</Tab>);
+        expect(wrapper.prop('aria-controls')).toBe(controlId);
+    });
+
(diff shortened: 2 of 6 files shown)


ivarni (approved):
Pro-tip, if you add "Fixes #626" to the commit-message the PR will automatically be linked to the issue and the issue will be closed once the PR is merged, I also believe the changelog will include a link to the issue.

ivarni (changes requested):
Actually, we have a policy of not renaming props, so really all you need to do here really is
```diff
Tab.propTypes = {
+ 'aria-controls': string.isRequired,
}
```

See https://github.com/SpareBank1/designsystem/blob/develop/CONTRIBUTING.md#ikke-rename-dom-props for details.

Sorry, I was a bit quick with the review earlier.

exoen:
> Actually, we have a policy of not renaming props, so really all you need to do here really is
> 
> ```diff
> Tab.propTypes = {
> + 'aria-controls': string.isRequired,
> }
> ```
> 
> See https://github.com/SpareBank1/designsystem/blob/develop/CONTRIBUTING.md#ikke-rename-dom-props for details.
> 
> Sorry, I was a bit quick with the review earlier.

I've addressed the comments and rebased. 

It says that merging can be automatically performed once the review has been addressed. It seems that GitHub does not pick up on update because of the rebase. Is rebasing not ideal when addressing PR reviews?

Anyhow, this can be merged now =)

ivarni (approved):

ivarni:
Honestly I've no idea how that automatic merging stuff on GitHub works but I personally prefer rebasing in almost all cases either way, since it keeps the history a lot cleaner.

Anyway, merging now. Thanks 👍

PR #629, opened 2019-05-13
Save space for error messages in order to avoid page jumps on validation

ivarni:
I decided to err on the side of caution and make the CSS change breaking because I think people who update to it should be aware of this change and hopefully a major version makes it more likely that they check the changelog. We're release so many minor/patch versions all the time that we can't really expect people to check the changelog for every new version.

For the react-component, no breaking change is needed at all since we're just adding a new property and none of the changes will take effect without updating the CSS package anyway.

### DIFF
--- a/packages/ffe-form-react/src/InputGroup.md
+++ b/packages/ffe-form-react/src/InputGroup.md
@@ -2,12 +2,29 @@ En _input group_ er en standarisert omverden for enkelt-inputs i et skjema. En g
 tooltip, skjemaelementet, og en feilmelding for valideringsfeil som vil bli vist om det er feil i skjemaet.
 
 ```js
-const { Input } = require('.');
+const { DatePicker } = require('../../ffe-datepicker-react');
+const { DropDown } = require('../../ffe-dropdown-react');
+const { Checkbox, Input, TextArea } = require('.');
+const { SecondaryButton } = require('../../ffe-buttons-react');
+
+initialState = { showErrors: false };
 
 <React.Fragment>
+    <div className="ffe-button-group">
+        <SecondaryButton
+            onClick={() =>
+                setState(prevState => ({
+                    showErrors: !prevState.showErrors,
+                }))
+            }
+        >
+            Skru feilmeldinger av/på
+        </SecondaryButton>
+    </div>
     <InputGroup
         label="Telefonnummer"
         tooltip="Vi bruker telefonnummer for å sende deg kvittering på SMS"
+        fieldMessage={state.showErrors ? 'Ugyldig telefonnummer' : null}
     >
         <Input
             type="tel"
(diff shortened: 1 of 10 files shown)


packages/ffe-form-react/src/index.d.ts
```diff
@@ -25,6 +25,7 @@ export interface BaseFieldMessageProps
     extends React.HTMLAttributes<HTMLDivElement> {
```
hdaljord: Here HTMLDivElement should be changed to HTMLElement as it is not known anymore that this is a div-element.

packages/ffe-form-react/src/InputGroup.md
```diff
@@ -2,12 +2,111 @@ En _input group_ er en standarisert omverden for enkelt-inputs i et skjema. En g
...
+        />
+    </InputGroup>
+
+    <InputGroup fieldMessage={state.showErrors ? 'Ooops' : null}>
+        <Checkbox name="check">Kryssboks</Checkbox>
+    </InputGroup>
+</React.Fragment>;
+```
+
+Default oppførsel er at det holdes av plass under inputfeltene for å vise en feilmelding uten at innholder lengre ned endres. Dersom man ikke ønsker dette kan det skrus
```
hdaljord: just a typo, should have been: ...innholdet lenger ned...

packages/ffe-form-react/src/RadioBlock.js
```diff
@@ -11,6 +11,12 @@ class RadioBlock extends Component {
             checked,
             children,
             className,
+            /* Support for the dark theme has not been added for this component
+             * but since we're passing props onwards to DOM nodes we need to make
+             * sure we don't pass this as part of "inputProps" since it's not being
+             * used but can still be passed down from a parent `RadioButtonInputGroup`.
+             */
+            dark, //eslint-disable-line
```
hdaljord, line 19: This should then be added as an optional variable to ts definitions, otherwise it will give an error.
If we want to pass everything on, we can use     [propName: string]: any;, but I would prefer to just add the parameter with boolean and add the same comment together with the type definition

packages/ffe-form-react/src/RadioButtonInputGroup.md
```diff
@@ -4,37 +4,227 @@ være et spørsmål:
...
+                >
+                    Vil ikke!
+                </RadioBlock>
+            </React.Fragment>
+        )}
+    </RadioButtonInputGroup>
+</React.Fragment>;
+```
+
+Default oppførsel er at det holdes av plass under inputfeltene for å vise en feilmelding uten at innholder lengre ned endres. Dersom man ikke ønsker dette kan det skrus
```
hdaljord: Nitpicking, same typo

packages/ffe-form-react/src/RadioBlock.js
ivarni: Ah ok. It makes sense that it would throw, since that's what passing it onwards does (React sees an all-lowercase prop and assumes it's meant as a custom attribute but those can only be strings), I just totally forgot that it's not something we're passing manually. Will fix.

packages/ffe-form-react/src/RadioButtonInputGroup.md
ivarni: Same typo? That's strange 😎

ivarni:
PR updated with review feedback

hdaljord (dismissed):

antidecaf (dismissed):
Nice one, thank you! :sunglasses:

ivarni:
@antidecaf  Updated `peerDeps` now since lerna seems to handle running `npm install` even with non-existing versions in peer-deps. I'm not sure if it's because of the `||` [here](https://github.com/SpareBank1/designsystem/blob/c5621a640ddf4ae4b6a2ce630f4c4ca285928480/packages/ffe-form-react/package.json#L51) that it works?

antidecaf (approved):

PR #632, opened 2019-05-13
Vertikal description list 

pethel:
Vi har behov for en vertikal description list. 

image

### DIFF
--- a/packages/ffe-lists-react/src/DescriptionList.js
+++ b/packages/ffe-lists-react/src/DescriptionList.js
@@ -2,10 +2,11 @@ import React from 'react';
 import { bool, string, node } from 'prop-types';
 import classNames from 'classnames';
 
-const DescriptionList = ({ medium, large, className, ...rest }) => (
+const DescriptionList = ({ medium, large, className, horizontal, ...rest }) => (
     <dl
         className={classNames(
             'ffe-description-list',
+            { 'ffe-description-list--horizontal': horizontal },
             { 'ffe-description-list--md': medium },
             { 'ffe-description-list--lg': large },
             className,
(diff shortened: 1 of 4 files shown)


packages/ffe-lists-react/src/DescriptionList.js
```diff
@@ -23,6 +24,8 @@ DescriptionList.propTypes = {
     medium: bool,
     /** Large list. Can't be used with `medium` prop */
     large: bool,
+    /** Displays dts and dds inline*/
+    horizontal: bool,
```
hdaljord, line 28: Please add this to the index.d.ts file as well (with a ?, since it's optional)

pethel: https://github.com/SpareBank1/designsystem/pull/632/commits/edca781aeb0019122ca5c1bdb53f41a97c0d1c3e

henrikhermansen (dismissed):
At first it seemed strange to make a breaking change to accommodate these needs.

However, on second thought, to not make a breaking change here would require us to add styling which would essentially be reverting the styling we already have. This breaking change makes a lot more sense.

I approve the changes, but it would be great if you could also update the markdown files with some relevant examples, so it will show on design.sparebank1.no.

henrikhermansen (dismissed):

antidecaf (dismissed):

antidecaf (approved):

ivarni:
Why can't you make the existing view the default so it doesn't have to be a breaking change? As far as I can see there was no reason for this change to be breaking. That creates unnessescary churn in my opinion. 

There's also nothing breaking about the change in the React component. Updating that on its own does not break anything and the changelog should mention that updating the CSS module is required for this new feature.

ivarni:
[peerDependencies](https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-lists-react/package.json#L44) should also be updated since the change was made breaking.

PR #636, opened 2019-05-14
fix(ffe-form-react): Fix tooltip bug in RadioButtonInputGroup

### LINKED ISSUE #631: InputGroup and RadioButtonInputGroup treat the tooltip prop different

`ffe-form-react`

Both `InputGroup` and `RadioButtonInputGroup` accepts `tooltip` props as both `string` and `node` but they treat them Differently.

The [InputGroup](https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-form-react/src/InputGroup.js#L74-L80) wraps `strings` in `<Tooltip>`, just renders an empty `Tooltip` for `booleans` and uses `React.isValidElement` otherwise to catch the case where the `tooltip` is a `node`. *This means the consumer must themselves ensure that the tooltip they are providing as a `node` is properly wrapped in the `<Tooltip>` element.*

The [RadioButtonInputGroup](https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-form-react/src/RadioButtonInputGroup.js#L7-L14) tries to be more clever and detect if the provided `tooltip` is already a `<Tooltip>` in which case it returns it as-is and otherwise it wraps it in `<Tooltip>`.  While this approach is definetely more elegant it sadly does not seem to work in newer versions of React (it seems `input.type` is no longer a thing) and if the provided tooltip is wrapped inside a `<Tooltip>` component it gets wrapped inside a second one. *This means the consumer must NOT wrap the tooltip they are providing in a `<Tooltip>` element.*

I would expect the two different flavours of input-group to behave in the same way, and I suspect this is a regression and that they used to work similar in earlier versions of React.

When providing a `tooltip` to `InputGroup` you must wrap in inside an instance of `<Tooltip>` while when providing a `tooltip` to `RadioButtoniInputGroup` you must not.

I suggest we copy the behaviour from `InputGroup` over to `RadioButtonInputGroup` and get rid of the `ensureComponent` helper completely. It's probably doing something similar for the `fieldMessage` prop.

ivarni:
When someone provided a `Tooltip` component as the `tooltip` prop to `RadioButtonInputGroup`, it
would be wrapped inside another `Tooltip` instance. It now acts the same way as the same prop in
`InputGroup`.

Fixes #631

### DIFF
--- a/packages/ffe-form-react/src/RadioButtonInputGroup.js
+++ b/packages/ffe-form-react/src/RadioButtonInputGroup.js
@@ -4,15 +4,6 @@ import classNames from 'classnames';
 import ErrorFieldMessage from './ErrorFieldMessage';
 import Tooltip from './Tooltip';
 
-const ensureComponent = Comp => (input, dark) => {
-    if (input.type === Comp) {
-        return input;
-    }
-    return <Comp dark={dark}>{input}</Comp>;
-};
-
-const ensureTooltip = ensureComponent(Tooltip);
-
 const RadioButtonInputGroup = props => {
     const {
         'aria-invalid': ariaInvalid,
@@ -37,8 +28,6 @@ const RadioButtonInputGroup = props => {
         selectedValue,
     };
 
-    const tooltipContent = tooltip && ensureTooltip(tooltip, dark);
-
     return (
         <fieldset
             className={classNames(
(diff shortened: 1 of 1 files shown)


hdaljord (dismissed):

ivarni:
Conflicts. Conflicts everywhere.

gbrownlee (approved):
I prefer this approach in part because it is the simplest. 👍 Also, I see no reason to get the entire tooltip-component from outside the inputGroup when it requires no config besides the text.

PR #651, opened 2019-06-12
fix(ffe-lists): no space when to dd folow each other

pethel:
Looks strange with a lot of padding when 2 or more `<dd/>` follow each other. There is no prev-sibling selector to remove the padding so my options are limited if I dont do a bigger rewrite. 

screencapture-localhost-6060-2019-06-12-14_18_05


ivarni (approved):
Never used this component but I don't see a problem with this and noone else has spoken up

antidecaf (approved):
LGTM!

Issue #655, opened 2019-06-27
Possibility to pass in your own id for child Input in InputGroup


hdaljord:
#### Which package is this issue related to?
ffe-form-react
#### Describe your issue (screenshots welcome!)
Today it is not possible to code your own id for the Input in an InputGroup, it will always be a unique generated id InputGroup.js, which gives a new id each time. 

#### What is the expected behavior? 
It should be possible to pass in your own id that should be used instead of the generated one when it is provided. This will allow the developer to have more control of the code and allow autocomplete on the input field.

![image](https://user-images.githubusercontent.com/36953762/60245168-4d51f080-98bc-11e9-8da5-a9b6cf47f364.png)

hdaljord:
Closed by #656 .

PR #656, opened 2019-06-27
Adding childId as an optional prop for InputGroup to be used as id on the Input child element

hdaljord:
Solves #655 

The childId property will only be used when it is provided, when it is not provided the component will work exactly as it did before.

This will allow the developer to have more control of the code and allow autocomplete on the input field.

### DIFF
--- a/packages/ffe-form-react/src/InputGroup.js
+++ b/packages/ffe-form-react/src/InputGroup.js
@@ -10,11 +10,13 @@ class InputGroup extends Component {
     constructor(props) {
         super();
 
-        this.id = `input-${uuid.v4()}`;
+        this.id = props.inputId ? props.inputId : `input-${uuid.v4()}`;
     }
 
     render() {
         const {
+            // eslint-disable-next-line no-unused-vars
+            inputId,
             children,
             className,
             extraMargin,
@@ -108,6 +110,8 @@ class InputGroup extends Component {
 const instanceOfComponent = component => shape({ type: oneOf([component]) });
 
 InputGroup.propTypes = {
+    /** The id that will be used on the input child if you don't want a generated one */
+    inputId: string,
     /** Unless you only have one element in your `InputGroup` you will have to use the function-as-a-child pattern. */
     children: oneOfType([func, node]).isRequired,
     className: string,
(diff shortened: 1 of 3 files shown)


packages/ffe-form-react/src/InputGroup.js
```diff
@@ -26,6 +27,8 @@ class InputGroup extends Component {
             ...rest
         } = this.props;
 
+        const id = childId ? childId : this.generatedId;
```
ivarni: You've already assigned `childId` to `this.generatedId` in the constructor if it is set, are we sure we want to allow consumers to change a field's ID after it has been mounted?

packages/ffe-form-react/src/InputGroup.js
```diff
@@ -108,6 +111,8 @@ class InputGroup extends Component {
 const instanceOfComponent = component => shape({ type: oneOf([component]) });
 
 InputGroup.propTypes = {
+    /** The id that will be used on the input child if you don't want a generated one */
+    childId: string,
```
ivarni: Maybe we can call it `inputId` since we're going to use it on the input-control? An `<InputGroup />` has several children, including labels, tooltips etc.

hdaljord: I agree, that's a much better name, I'll fix

packages/ffe-form-react/src/InputGroup.js
hdaljord: No, that was not my intention. I just wanted the childId not to appear in the generated div element, if it is not present in line 18 it will be passed along with the ...rest. 

I see that I can achieve the same by adding an eslint-ignore, do you think that's a better solution, or do you have another suggestion?

ivarni: I'd just add an ignore in this case. I think it's even possible to configure eslint to not complain about unused variables when there's a `...rest` present but I can't remember how to do it.

Either way is fine with me, I just wanted to know if there was an actual usecase I wasn't seeing :) Techincally any ID we can be changed at any time so it's not really a big issue if this one can as well.

hdaljord: Updated

ivarni (approved):

PR #657, opened 2019-06-27
Avoid circular imports in grid package

jarlef:
Avoid importing InlineGrid into utils.dev.js which results in circular import chain. The only purpose is of the import is to do a type check. A simpler approach is to compare  child.type.name. against a whitelist registry of allowed nested components instead of comparing react functions.


ivarni (approved):
LGTM 👍

PR #663, opened 2019-07-16
feat(ffe-datepicker-react): Keep display state on validation error

Snikanes:
We have a use-case where we use two Datepickers in order to allow selection of a date range. In this use-case the `minDate` of the end-datepicker is given by the value of the start-datepicker and vice versa. Selecting a value in one of the datepickers may causes an invalid minDate/maxDate combination in the other, causing the datepicker with the now invalid combination to display its calendar. In this PR I have added a prop that allows us to override this behavior, keeping the value of `displayDatepicker` as is when such an invalid combination leads to a validation error.

### DIFF
--- a/packages/ffe-datepicker-react/src/datepicker/Datepicker.js
+++ b/packages/ffe-datepicker-react/src/datepicker/Datepicker.js
@@ -87,7 +87,12 @@ export default class Datepicker extends Component {
 
     validateDateIntervals() {
         let nextState = {};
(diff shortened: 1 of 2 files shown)


ivarni (approved):
I've never really worked with this codebase or used the component but this doesn't look like it breaks anything to me so in the interest of keeping things moving during summer I'm 👍

Snikanes:
The added functionality here is really opt-in, so I think we should be good.

PR #665, opened 2019-07-18
Spacing and typo adjustments

antidecaf:
* Added new variables for standardized spacing
* Adjusted typography according to new values in the Figma master file

### DIFF
--- a/packages/ffe-core/less/ffe.less
+++ b/packages/ffe-core/less/ffe.less
@@ -2,6 +2,7 @@
 @import 'ffe-normalize';
 @import 'colors';
 @import 'font-sizes';
+@import 'spacing';
 @import 'typography';
 @import 'layout';
 @import 'motion';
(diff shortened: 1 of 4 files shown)


ivarni:
If we're introducing a whole new system for spacing, doesn't that justify a new major version? 
Font-size changes, especially increases, has a non-zero chance of breaking existing layout.

Some instructions on how to use these variables and general information about the new spacing system would also be a very nice thing.

antidecaf:
> Font-size changes, especially increases, has a non-zero chance of breaking existing layout.

Yeah, agreed. I'll rewrite that commit message to include breaking change.

> If we're introducing a whole new system for spacing, doesn't that justify a new major version?

> Some instructions on how to use these variables and general information about the new spacing system would also be a very nice thing.

I've only added the variables for now in order make them available in different apps, will document and implement them in the design system components in a later PR.

ivarni:
Sounds good. Looking forward to when this is all in place. We have a lot of inconsistent spacing around.

antidecaf:
Indeed we do. And that makes the upcoming PR a bit painful. :sob:

ivarni (approved):

PR #666, opened 2019-07-26
feat(ffe-context-message-react): ContextMessage components now pass r…

ivarni:
…est props onto root node

### DIFF
--- a/packages/ffe-context-message-react/src/ContextMessage.js
+++ b/packages/ffe-context-message-react/src/ContextMessage.js
@@ -73,6 +73,7 @@ export default class ContextMessage extends Component {
             messageType,
             showCloseButton,
             style,
+            ...rest
         } = this.props;
 
         if (this.state.closed) {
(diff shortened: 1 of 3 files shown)


henrikhermansen:
I'd like to suggest changing the commit message. Of course that long scope doesn't make it easy, but how about this?
> feat(ffe-context-message-react): pass rest props to root node
>
> ContextMessage components now pass rest props onto root node.

ivarni:
@henrikhermansen Sure. Not sure it makes the changelog any more readable but I'd kinda just want this merged so I've updated the commit message :)

hdaljord (changes requested):
Can you please add `extends React.HTMLAttributes<HTMLDivElement>` i index.d.ts for ContextMessageProps, to make this work for typescript users as well.

Like this:
`export interface ContextMessageProps extends React.HTMLAttributes<HTMLDivElement>{
`

ivarni:
@hdaljord 

Done. I've no way of testing if it works though :)
~~Shittier~~ Prettier decided to automatically add a linebreak there but I assume that doesn't matter.

hdaljord:
Nice, I've tested it locally :)

hdaljord (approved):

PR #667, opened 2019-08-01
feat(ffe-icons): New icons (drafts)

jarlef:
Added new icons from based on export from the sketch file of the design system which contains alot more icons than that exists in the current design system

Icons which is prefixed with 'draft-' is shown in a separate page (Nye ikoner (UTKAST)). This page is only displayed in dev mode.


ivarni (approved):
I'd like the fact that they're not available for use yet to be included in the commit message so it's also included in the changelog, otherwise 👍

antidecaf (approved):

PR #668, opened 2019-08-01
feat(ffe-icons): Added import script

jarlef:
Automatically import svg icons using svgo and svg-crop to optimize the svg. The newly imported files is imported with a `draft-` prefix. See #667 for handling of draft icons.

### DIFF
--- a/packages/ffe-icons/bin/import.js
+++ b/packages/ffe-icons/bin/import.js
@@ -1,0 +1,74 @@
+#!/usr/bin/env node
+const fs = require('fs');
+const path = require('path');
+
+const optimizer = require('./optimizer');
+const autoCrop = require('svg-autocrop');
+
+const importPath = path.join(__dirname, '../import');
+const destPath = path.join(__dirname, '../icons');
+
+const crop = true;
+
+const log = console.log;
+
+const normalizeName = fileName => {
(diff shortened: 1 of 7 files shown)


packages/ffe-icons/bin/import.js
```diff
@@ -0,0 +1,74 @@
...
+    const destPathExists = fs.existsSync(destPath);
+    if (!importFolderExists) {
+        throw new Error(`Import folder ($(importPath)) not found`);
+    }
+
+    if (!destPathExists) {
+        fs.mkdirSync(destPath);
+    }
+
+    const files = await fs.readdirSync(importPath, { encoding: 'utf-8' });
```
ivarni, line 42: I don't think there's much point `await`ing a synchronous function, it'll just get transpiled down to an immediately resolved promise.

ivarni (commented):
Should we also remove the old sections of the docs regarding the importing of icons?

https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-icons/README.md#adding-a-new-icon-to-ffe-icons

https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-icons/icons/_ICON_REQUIREMENTS.md

It seems this script is meant to handle what used to be a manual process?

PR #671, opened 2019-08-08
fix(ffe-form-react): Fixing typescript type definition for InputGroup…

### LINKED ISSUE #670: Feil i type for tooltip på InputGroup
ffe-form-react

tooltip er definert som 
{code}
tooltip?: boolean | string | Tooltip;
{code}

Skulle vært definert som 

{code}
tooltip?: boolean | string | React.ReactNode;
{code}

Forårsaker en feilmelding som gjør det umulig å bruke tooltip til <Tooltip/>.

hdaljord:
…Props.tooltip to make it possible to send in a Tooltip-component.

This PR fixes #670. The change is tested and verified in a project locally.

### DIFF
--- a/packages/ffe-form-react/src/index.d.ts
+++ b/packages/ffe-form-react/src/index.d.ts
@@ -63,14 +63,14 @@ export interface LabelProps
 export interface InputGroupProps extends React.HTMLAttributes<HTMLDivElement> {
     inputId?: string;
     /** Unless you only have one element in your `InputGroup` you will have to use the function-as-a-child pattern. */
-    children: JSX.Element;
+    children: React.ReactNode;
     className?: string;
     extraMargin?: boolean;
     fieldMessage?: string | React.ReactNode;
     description?: string;
     label?: string | Label;
     onTooltipToggle?: (e: React.MouseEvent | undefined) => void;
-    tooltip?: boolean | string | Tooltip;
+    tooltip?: React.ReactNode;
 }
 
 export interface TooltipProps extends React.HTMLAttributes<HTMLSpanElement> {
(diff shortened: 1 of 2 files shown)


packages/ffe-form-react/src/index.d.ts
```diff
@@ -70,7 +70,7 @@ export interface InputGroupProps extends React.HTMLAttributes<HTMLDivElement> {
     description?: string;
     label?: string | Label;
     onTooltipToggle?: (e: React.MouseEvent | undefined) => void;
-    tooltip?: boolean | string | Tooltip;
+    tooltip?: boolean | string | React.ReactNode;
```
ivarni: Isn't [booleans and strings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L186-L191) also `ReactNode`s? I know it works like that with regular prop-types and it seems that way in the definition of `ReactNode` as well, if I understand the syntax.

If that's the case it seems it should be enough to just say `tooltip?: React.ReactNode`.

By the way, the `instanceOfComponent` check in the propType declaration is also wrong, it should just have been `tooltip: node` [here](https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-form-react/src/InputGroup.js#L131). It's throwing errors currently and I don't think it's worked for the last few versions of React. Since you're already here that would be a nice thing to fix as well :)

hdaljord: Yes, that's true. I'll fix both :)

ivarni (dismissed):

ivarni (approved):

PR #682, opened 2019-08-27
chore: bump React to 16.9 to allow for using hooks etc

henrikhermansen:
BREAKING CHANGE: peerDep of React is raised to 16.9. While this in itself
isn't a breaking change, it allows for the use of hooks in later releases.
Such changes would just be a minor version, or even a patch, but without
this major version bump it would result in a runtime error for consumers.


henrikhermansen:
Does anyone know if it's problematic to use `chore` commit for this kind of change?
Will it appear in changelogs and bump the version correctly when there's a breaking change present?

antidecaf:
@henrikhermansen Breaking changes can occur in any type of commit, according to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/).

ivarni:
I think BREAKING CHANGE in the commit-message trumps whatever you use for the commit tag.

But maybe we should go wild and require 16.9.0 while we're at it? The profiler API might be handy.

antidecaf (dismissed):

henrikhermansen:
> I think BREAKING CHANGE in the commit-message trumps whatever you use for the commit tag.
> 
> But maybe we should go wild and require 16.9.0 while we're at it? The profiler API might be handy.

I wouldn't mind. If someone hasn't upgraded to 16.8 yet, it won't matter to them whether they upgrade to 16.8 or 16.9.
Would you mind if I go ahead and make it 16.9 instead @antidecaf ?

ivarni (approved):

antidecaf (approved):

PR #683, opened 2019-08-27
Collapse component

henrikhermansen:
Add new package with a collapse component to replace the outdated react-css-collapse.

This component is now used in ffe-accordion-react and ffe-form-react.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -22,6 +22,7 @@ module.exports = {
                 'ffe-context-message-react',
                 'ffe-core',
                 'ffe-core-react',
+                'ffe-collapse-react',
                 'ffe-datepicker',
                 'ffe-datepicker-react',
                 'ffe-decorators-react',
--- a/packages/ffe-accordion-react/package.json
+++ b/packages/ffe-accordion-react/package.json
@@ -26,10 +26,10 @@
     "setupTestFrameworkScriptFile": "../../test-setup.js"
   },
   "dependencies": {
+    "@sb1/ffe-collapse-react": "^0.0.0",
     "@sb1/ffe-icons-react": "^7.0.0",
     "classnames": "^2.2.5",
     "prop-types": "^15.6.0",
-    "react-css-collapse": "^3.6.0",
     "uuid": "^3.1.0"
   },
   "devDependencies": {
--- a/packages/ffe-accordion-react/src/AccordionItem.js
+++ b/packages/ffe-accordion-react/src/AccordionItem.js
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import { arrayOf, bool, func, node, number, string } from 'prop-types';
-import Collapse from 'react-css-collapse';
+import Collapse from '@sb1/ffe-collapse-react';
 import ChevronIkon from '@sb1/ffe-icons-react/lib/chevron-ikon';
 import classNames from 'classnames';
 
--- a/packages/ffe-all.less
+++ b/packages/ffe-all.less
@@ -19,3 +19,4 @@
 @import './ffe-system-message/less/ffe-system-message';
 @import './ffe-tables/less/tables';
 @import './ffe-tabs/less/tabs';
+@import './ffe-collapse-react/less/collapse';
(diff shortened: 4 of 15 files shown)


packages/ffe-accordion/less/ffe-accordion.less
```diff
@@ -110,17 +110,7 @@
...
-    will-change: height;
-    height: 0;
-    overflow: hidden;
-    visibility: hidden;
     transition: height @ffe-transition-duration @ffe-ease-in-out-back;
-
-    .ffe-accordion-item--open & {
-        height: auto;
-        overflow: visible;
-        visibility: visible;
```
henrikhermansen: As I make the PR I'm thinking: what about those who just use the styling, and not the react packages?

I suppose I should perhaps revert this? The same goes for ffe-form.less.

henrikhermansen:
I'm unsure about releasing new packages this way. When I checked with `lerna version` it seems it will release `1.1.0` of `ffe-css-collapse-react`. How do I make it release with `1.0.0`?

packages/ffe-accordion/less/ffe-accordion.less
ivarni: Ideally yes, though I can't really say I see much involvement from the "non react" teams we still have the static pages from the CMS at sparebank1.no

ivarni:
Any chance we can just call it `ffe-collapse-react`? The previous package was named `react-css-collapse` because the name `react-collapse` was taken. We don't have that problem.

packages/ffe-css-collapse-react/less/css-collapse.less
```diff
@@ -0,0 +1,3 @@
+.ffe-css-collapse-transition {
+    transition: height @ffe-transition-duration @ffe-ease-in-out-back;
```
ivarni: Maybe we could just inline this style and avoid having a `.less` file at all?
If not then the readme should probably mention that it exists and needs to be imported unless you're importing the `ffe-all.less` file (which was new to me).

henrikhermansen: I thought about it, but what about the less variables? I considered it nice that this component would follow if we change our default duration or easing.

I would rather go for the update of the readme.

ivarni: Fair enough, I forgot about the variable :)

packages/ffe-accordion/less/ffe-accordion.less
henrikhermansen: Allright, I removed the commits that removes the styling.

ivarni:
Lerna is being derpy again
```
lerna ERR! npm install exited 1 in '@sb1/ffe'
lerna ERR! npm install stderr:
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/@sb1%2fffe-css-collapse-react - Not found
npm ERR! 404 
npm ERR! 404  '@sb1/ffe-css-collapse-react@^1.0.0' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 It was specified as a dependency of 'designsystem'
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```
I think you'll have to split this into two PRs where the first one publishes the new package and the next one makes it being used by the other components 🤡

henrikhermansen:
> Any chance we can just call it `ffe-collapse-react`? The previous package was named `react-css-collapse` because the name `react-collapse` was taken. We don't have that problem.

Yeah, that's true, it's not of significance that we use css for the transition. I've renamed everything.

henrikhermansen:
> Lerna is being derpy again
>
> I think you'll have to split this into two PRs where the first one publishes the new package and the next one makes it being used by the other components

I just pushed again. If this doesn't work now, I'll split it into two PRs :D

packages/ffe-css-collapse-react/less/css-collapse.less
henrikhermansen: I updated the readme and the styleguide-content document.

ivarni (dismissed):

antidecaf (dismissed):

antidecaf (approved):

Issue #132, opened 2018-03-28
Improve the ffe-tables-react API


selbekk:
#### Which package is this issue related to?
ffe-tables-react 

#### Describe your issue (screenshots welcome!)
The API is massive, and hard to use. Several people in our team avoid this component, just because the level of abstraction is too damn high.

#### What is the expected behavior? 
I would love an API that's less abstracted, and that's discussed thoroughly here before implemented.

This is an issue for discussing, suggesting alternative APIs and considering the pros and cons of each.

ksawerysb1:
Add a columns.render function as an alternative to defining custom component fields in the data object rows.

wkillerud:
The SortableTable component is perhaps beyond saving :P Perhaps we can use render functions to move more control from configs to the render function?

For `Table` I think we can go back to basics:

```
<Table caption="Utviklere">
    <TableHead>
        <TableHeader>
            Navn
        </TableHeader>
        <TableHeader>
            Alder
        </TableHeader>
    </TableHead>
    <TableBody>
        <TableRow>
            <TableCell column="Navn">
                Anders
            </TableCell>
            <TableCell column="Alder">
                32
            </TableCell>
        </TableRow>
    </TableBody>
</Table>
```

```
const data = [
    { name: 'Anders', age: 32 },
    { name: 'Erik', age: 25 },
    { name: 'Gunn', age: 29 },
    { name: 'Safi', age: 42 },
    { name: 'Sandra', age: 42 },
];

<Table caption="Utviklere">
    <TableHead>
        <TableHeader>
            Navn
        </TableHeader>
        <TableHeader>
            Alder
        </TableHeader>
    </TableHead>
    <TableBody>
        {data.map(row => (
            <TableRow>
                <TableCell column="Navn">
                    {row.name}
                </TableCell>
                <TableCell column="Alder">
                    {row.age}
                </TableCell>
            </TableRow>
        ))}
    </TableBody>
</Table>
```

wkillerud:
With the API from the previous comment making a SortableTable could be a matter of adding a few props to `TableHeader` and defer sorting to the consumer.

```
const data = [
    { name: 'Anders', age: 32 },
    { name: 'Erik', age: 25 },
    { name: 'Gunn', age: 29 },
    { name: 'Safi', age: 42 },
    { name: 'Sandra', age: 42 },
];

<Table caption="Utviklere">
    <TableHead>
        <TableHeader
[...]

PR #688, opened 2019-08-29
Remove deprecated lifecycles

### LINKED ISSUE #372: Future-proof React 17 components

henrikhermansen:
This PR removes the final deprecated React lifecycle from our components.

While the issue #372 also mentions `ffe-tables-react` I can't find any deprecated methods there. However, there is made use of `getDerivedStateFromProps` which in most cases can (and should) be replaced or rewritten. I think a separate issue should be raised for this component, or it could be solved by solving #132 

The PR also adds rules for hooks to our eslint config.

Fixes #372

### DIFF
--- a/packages/ffe-account-selector-react/src/subcomponents/input-field/InputField.js
+++ b/packages/ffe-account-selector-react/src/subcomponents/input-field/InputField.js
@@ -1,147 +1,116 @@
 /* Because we are missing aria-controls (http://www.heydonworks.com/article/aria-controls-is-poop): */
 /* eslint jsx-a11y/role-has-required-aria-props:0 */
-import React, { Component } from 'react';
+import React from 'react';
 import { func, string, bool, number } from 'prop-types';
-import autoBind from 'react-auto-bind';
-import KryssIkon from '@sb1/ffe-icons-react/lib/kryss-ikon';
-import ChevronIkon from '@sb1/ffe-icons-react/lib/chevron-ikon';
+import { KryssIkon, ChevronIkon } from '@sb1/ffe-icons-react';
 import classNames from 'classnames';
 import txt from '../../i18n/i18n';
 import { Locale } from '../../util/types';
 
-class Input extends Component {
-    constructor(props) {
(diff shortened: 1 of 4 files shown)


ivarni (commented):
LGTM but I'd recommend getting someone who is using the component to take a look as well

ivarni (dismissed):

packages/ffe-account-selector-react/src/subcomponents/input-field/InputField.js
```diff
@@ -1,147 +1,114 @@
...
-        this.props.onFocus();
-    }
-
-    onBlur() {
-        this.setState({ isFocused: false });
-        this.props.onBlur();
-    }
-
-    onReset(e) {
+    const onReset = e => {
```
pethel: onReset -> handleReset?
Blir vell samma med de on... andra hvis du har lust.

henrikhermansen: Yup, that should be fine. This way I can deconstruct everything from props.

It does mean you have to approve the PR again though ;)

pethel (approved):

PR #689, opened 2019-08-29
fix(ffe-form-react): Added generated id on tooltip button to avoid …

hdaljord:
…warnings when used in forms validated with formik (expects id on every input or button field).

Formik exptects all input elements to have an id or name, so when tooltips are used inside forms validated with formik it outputs a warning when leaving the tooltip-button (keyboard navigation).

This fix adds a generated id to avoid this warning.
 
image

### DIFF
--- a/packages/ffe-form-react/src/Tooltip.js
+++ b/packages/ffe-form-react/src/Tooltip.js
@@ -53,6 +53,7 @@ class Tooltip extends React.Component {
                     onClick={this.onToggle}
                     type="button"
                     tabIndex={tabIndex}
+                    id={uuid.v4()}
                 >
                     <span aria-hidden={true}>?</span>
                 </button>


antidecaf (approved):

packages/ffe-form-react/src/Tooltip.js
```diff
@@ -53,6 +53,7 @@ class Tooltip extends React.Component {
                     onClick={this.onToggle}
                     type="button"
                     tabIndex={tabIndex}
+                    id={uuid.v4()}
```
ivarni, line 56: Maybe we can generate the id in the constructor like we already do with `tooltipId`?

This will generate a new id on every single render  and trigger unnecessary DOM-updates. While running the render method is quite cheap, touching the DOM isn't.

hdaljord: That's a good point, I'll create a new fix for that, @ivarni

antidecaf: Seems I may have been a bit trigger happy, sorry about that.

PR #690, opened 2019-08-29
fix(ffe-form-react): Moved tooltip button id generation to constructo…

hdaljord:
…r to avoid unnecessary DOM-updates.
 
Fixes @ivarni's comment on PR #689.


henrikhermansen:
The changes looks good, but the commit message is a bit weird.
The description should be a quick summary, and it's usually not started with a lower case. Then you may elaborate in the optional body.
How about something like this?
```
fix(ffe-form-react): move id generation to constructor

Moved tooltip button id generation to constructor to avoid unnecessary
DOM-updates.
```

ivarni (approved):

PR #692, opened 2019-09-06
feat(ffe-buttons-react): accept refs created by useRef

henrikhermansen:
The prop-type `innerRef` now accepts objects, in addition to functions.
With hooks it's now possible to create refs using the `useRef` hook.
These refs are objects which can be passed directly, instead of using a
ref-creating function.

### DIFF
--- a/packages/ffe-buttons-react/src/ActionButton.js
+++ b/packages/ffe-buttons-react/src/ActionButton.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { bool, func, node, string, oneOfType } from 'prop-types';
+import { bool, func, node, string, oneOfType, object, shape } from 'prop-types';
 import classNames from 'classnames';
 
 import Button from './BaseButton';
@@ -38,8 +38,8 @@ ActionButton.propTypes = {
     element: oneOfType([func, string]),
     /** Applies the ghost modifier if true. */
     ghost: bool,
-    /** Ref-setting function passed to the button element */
-    innerRef: func,
+    /** Ref-setting function, or ref created by useRef, passed to the button element */
+    innerRef: oneOfType([func, shape({ current: object })]),
     /** Shows a loader if true */
     isLoading: bool,
     /** Icon shown to the left of the label */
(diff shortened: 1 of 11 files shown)


henrikhermansen:
Hang on.. come to think of it, I'll use `shape` to make sure the ref object it has a `current` prop.

henrikhermansen:
> Hang on.. come to think of it, I'll use `shape` to make sure the ref object it has a `current` prop.

Aaaand done. Ready to be reviewed!

hdaljord (commented):
Can you please give an example of what kind of objects you want to pass in?
I ask to be sure that the type definitions is up-to-date, if the ref could be any other than a HTMLElement we need to update the index.d.ts as well

henrikhermansen:
> Can you please give an example of what kind of objects you want to pass in?
> I ask to be sure that the type definitions is up-to-date, if the ref could be any other than a HTMLElement we need to update the index.d.ts as well

It's the object created by the React hook `useRef`. This object has a property `current` which will either be `undefined` or an object which is an instance of `Element`. I guess that's the same kind of element as TS' `HTMLElement`? I have no idea if this will match the ref object matcher we currently have in place.

As it is now (without these changes) `innerRef` expects a function. How does that match the TS definition of a ref object? Is that just "how it is" or is it because the ref creator function returns a ref object, and so it matches the TS definition? I'm just asking out of curiosity since I don't know exactly what the ref creator functions does, nor do I know TS :)

Edit: code from React
```
function createRef() {
  var refObject = {
    current: null
  };
  {
    Object.seal(refObject);
  }
  return refObject;
}
```
The only difference seems to be that the ref creator function sets `current` to `null`, whereas `useRef` will set `current` to `initialValue` which is `undefined` when you don't provide any. It would be strange if the TS ref object matcher doesn't match both.

ivarni:
You should never be passing HTML elements to a `ref` prop, you should be passing something that you expect React to *populate with* a HTML element. Since that happens runtime I'm not sure how TS should be able to handle how the `ref` prop is handled after rendering or why it should even care but then I've been negatively surprised by TS before.

To me this looks unproblematic under the assumption that TS isn't involved at runtime.

ivarni (approved):

hdaljord (approved):
Thanks for the explanation. This should be covered by the definiition

`    innerRef?: React.RefObject<HTMLElement>;`

It's a good question if the current definition is correct for a function, I have only used it with refs created with React.createRef, so it might not be correct. 

Anyhow, this PR does not break anything, I'll test the function case later and if it's not correct I'll create a new PR for that case.

PR #694, opened 2019-09-09
Feature/dig 53247 file upload frontend

jafeld:
Big redesign of ffe-file-upload-react including new styling package ffe-file-upload, new design can be seen here (requires access): https://www.figma.com/file/cn3VFs107XnyxeLo4Du4ny/Fil-Opplasting---Master?node-id=0%3A1

Some examples of the new design: 

ffe-file-upload-1

ffe-file-upload-2

These are breaking changes.


ivarni:
Travis doesn't seem happy, https://travis-ci.org/SpareBank1/designsystem/builds/582730542#L837, but other than that it looks good to me.

jafeld:
> Travis doesn't seem happy, https://travis-ci.org/SpareBank1/designsystem/builds/582730542#L837, but other than that it looks good to me.

Yeah, this is due to the role="presentation" I removed on the 'drag and drop' receiver element based on comment on the other PR. Unsure what is preferred here :)

ivarni:
Then I would suggest the [region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Region_role) role along with aria-label describing it as a droppable area.

ivarni (dismissed):

jafeld:
> Then I would suggest the [region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Region_role) role along with aria-label describing it as a droppable area.

Based on conversation in PM, we decided to add the role="presentation" back and we'll create new case issue to deal with potential screen-reader problems

packages/ffe-file-upload/less/file-upload.less
```diff
@@ -0,0 +1,337 @@
+@ffe-blue-royal-escape: escape('@{ffe-blue-royal}');
+@ffe-blue-royal-fill: @ffe-blue-royal-escape;
+@ffe-blue-azure-escape: escape('@{ffe-blue-azure}');
+@ffe-blue-azure-fill: @ffe-blue-azure-escape;
+
+.ffe-file-upload {
+    margin-top: 45px;
+    display: -ms-grid;
```
antidecaf: Vendor prefixes are added at build time, so no need to include them in the Less files (will also cause stylelint errors/warnings).

packages/ffe-file-upload/less/file-upload.less
```diff
@@ -0,0 +1,307 @@
...
+            bottom: 5px;
+
+            @media (min-width: @breakpoint-md) {
+                width: 270px;
+                height: 210px;
+                right: 0;
+                bottom: 0;
+            }
+
+            &__hover {
```
kwltrs: this should be a modifier, i.e. `--hover`

kwltrs: this should be a modifier, i.e. `--hover`

packages/ffe-file-upload/less/file-upload.less
```diff
@@ -0,0 +1,307 @@
...
+        grid-template-columns: 1fr;
+
+        &-info {
+            display: grid;
+            grid-column: 1;
+            grid-row: 1;
+            grid-template-columns: 1fr;
+            overflow: hidden;
+
+            &-background {
```
kwltrs: The level of nesting is rather confusing and hard to follow. Could you change the nesting so that element names are not nested? (i.e. `&__file-item-stencil-info-background` instead of `&__file-item-stencil &-info &-background`)

packages/ffe-file-upload/less/file-upload.less
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    &__file-item-error {
+        display: grid;
+        grid-gap: 5px;
+        grid-template-columns: 25px 1fr;
+
+        &-icon {
+            &:extend(.ffe-field-error-message);
+
```
kwltrs, line 278: The definition of `.ffe-field-error-message` is provided by `ffe-form`, but `ffe-form` is not listed as dependency in `package.json`.

packages/ffe-file-upload/less/file-upload.less
```diff
@@ -0,0 +1,307 @@
+@ffe-blue-royal-escape: escape('@{ffe-blue-royal}');
+@ffe-blue-royal-fill: @ffe-blue-royal-escape;
+@ffe-blue-azure-escape: escape('@{ffe-blue-azure}');
+@ffe-blue-azure-fill: @ffe-blue-azure-escape;
+
+.ffe-file-upload {
+    margin-top: 45px;
+    display: grid;
+    max-width: 400px;
+    grid-template-columns: 1fr;
```
kwltrs, line 10: Could you add a line to `README` regarding browser support as this package is using css grids?

packages/ffe-file-upload/less/file-upload.less
kwltrs: And out of curiosity, does the `:hover` pseudo selector not work here? Something to do with drag and drop?

kwltrs (changes requested):
Looks like a very useful component. Thanks!

packages/ffe-file-upload/less/file-upload.less
jafeld: > this should be a modifier, i.e. `--hover`

It is a modifier, eg: 'block__component__modifier'

This modifier is only triggered when dragging files over the component to trigger the modified style, and as such I had to make a custom modifier. Other components, eg the 'delete' button, triggers on mouse-over ( the :hover selector ).

packages/ffe-file-upload/less/file-upload.less
jafeld: > The level of nesting is rather confusing and hard to follow. Could you change the nesting so that element names are not nested? (i.e. `&__file-item-stencil-info-background` instead of `&__file-item-stencil &-info &-background`)

Sure :) I'll give it another look!

packages/ffe-file-upload/less/file-upload.less
jafeld: Nice catch!

packages/ffe-file-upload/less/file-upload.less
jafeld: Should be fixed now (otherwise Travis will complain anyway :smile:  )

packages/ffe-file-upload/less/file-upload.less
henrikhermansen: > It is a modifier, eg: 'block__component__modifier'

It should be `block__element--modifier` as described by [BEM](http://getbem.com/naming/).
Thanks for clearing up why `:hover` isn't suited here.

packages/ffe-file-upload/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "lint": "stylelint less/*.less",
+    "test": "npm run lint"
+  },
+  "devDependencies": {
+    "@sb1/ffe-core": "^14.0.4",
+    "stylelint": "^9.4.0"
+  },
+  "peerDependencies": {
+    "@sb1/ffe-core": "^14.0.2",
+    "@sb1/ffe-form": "^10.1.1"
```
henrikhermansen: I would assume this would work with `@sb1/ffe-core@15.0.0` and `@sb1/ffe-form@11.0.0` as well?
If so it should be reflected in the peerDependencies.

packages/ffe-file-upload-react/src/FileItem.js
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
+import { number, func, shape, object, string as stringType } from 'prop-types';
+
+const FileItem = props => (
```
henrikhermansen: Could you please deconstruct props here?
It's nice to be able to see available props right away, and you'll save yourself alot of `props.` throughout the code.

packages/ffe-file-upload-react/src/FileItem.js
```diff
@@ -0,0 +1,96 @@
...
+    <li key={props.index}>
```
henrikhermansen: I don't see a purpose of providing a key here.
If we are mapping an array to several `FileItem`s somewhere else we would rather give a key to each `FileItem`, as that's where React actually needs it.

packages/ffe-file-upload/less/file-upload.less
jafeld: > > It is a modifier, eg: 'block__component__modifier'
> 
> It should be `block__element--modifier` as described by [BEM](http://getbem.com/naming/).
> Thanks for clearing up why `:hover` isn't suited here.

Ah, right! My bad :)

packages/ffe-file-upload-react/src/FileItem.js
```diff
@@ -0,0 +1,96 @@
...
+        {// File is loading
+        !props.file.document && !props.file.error && (
+            <div className="ffe-file-upload__file-item-stencil">
```
henrikhermansen, line 8: There are alot of `div`s throughout this component.
I'm no expert on semantics, but could any of these `div`s possibly be something else to add semantics?

@ivarni

packages/ffe-file-upload-react/src/FileUpload.js
```diff
@@ -29,97 +33,103 @@ class FileUpload extends React.Component {
...
-                    className="ffe-file-upload__button"
-                    aria-invalid={String(!!errorMessage)}
-                    onClick={this.triggerUploadFileNativeHandler}
+                <div className="ffe-file-upload__title">{title}</div>
+                {Object.keys(files).length > 0 ? (
+                    <div className="ffe-file-upload__file-items-section">
+                        {Object.keys(files).map((file, index) => (
+                            <FileItem
+                                key={index}
+                                index={index}
```
henrikhermansen: `index` isn't needed here.
`key` is recommended to not use index from `.map()`, but rather some value which actually corresponds to the data you want to use. Perhaps simply using `file` here will do?

Edit: if `file` is the file name, we might get duplicates. You could try using `files[file]` instead.

packages/ffe-file-upload-react/src/FileUpload.js
```diff
@@ -29,97 +33,103 @@ class FileUpload extends React.Component {
...
                 >
-                    <span className="ffe-file-upload__button__label-icon" />
-                    <span id={`${id}-label`}>{label}</span>
-                </button>
+                    <div
+                        className={`ffe-file-upload__upload-section-border${
+                            this.state.hover
+                                ? ' ffe-file-upload__upload-section-border__hover'
+                                : ''
+                        }`}
```
henrikhermansen: You could rather use the `classnames` package here.

```
className={ classnames(
  'ffe-file-upload__upload-section-border',
  { 'ffe-file-upload__upload-section-border__hover': this.state.hover }
) }
```

The package is used alot throughout Designsystem.

henrikhermansen (changes requested):

packages/ffe-file-upload-react/src/FileItem.js
ivarni: I'm not familiar with the correct semantics for file uploads but we discussed the accesibility part of it and decided we'd make a separate issue for that. Since the upload button is inside the drag-and-drop target container it's difficult to pull that off without just providing a second upload button that is hidden for non-screenreaders outside of it and then just hiding everything else for screenreaders. At that point semantics matter a bit less.

But I didn't really want to mix that into this redesign since this PR is already quite big.

packages/ffe-file-upload-react/src/FileUpload.js
ivarni: There's no issue with using the index as a key as long as you don't plan on changing the sort order.

ivarni: Hm.. well actually in this case that can happen since there is functionality to delete a file. If one is deleted from the front of the list then all the other ones will have their index changed so in this case it actually does matter.

I'll go to my corner.

jafeld: > `index` isn't needed here.
> `key` is recommended to not use index from `.map()`, but rather some value which actually corresponds to the data you want to use. Perhaps simply using `file` here will do?
> 
> Edit: if `file` is the file name, we might get duplicates. You could try using `files[file]` instead.

File is an object, so even if we have duplicates, they should not be equal, right?

ivarni (dismissed):

ivarni:
@henrikhermansen @kwltrs your requested changes is what's stopping the merge right now. I'm not sure how to deal with them.

kwltrs (dismissed):

antidecaf (dismissed):

packages/ffe-file-upload/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "repository": {
+    "type": "git",
+    "url": "ssh://git@github.com:SpareBank1/designsystem.git"
+  },
+  "scripts": {
+    "lint": "stylelint less/*.less",
+    "test": "npm run lint"
+  },
+  "devDependencies": {
+    "@sb1/ffe-core": "^14.0.4",
```
henrikhermansen: This is now mismatching your peer dependency. I would suggest perhaps accepting a wider range of versions in peer dependencies.
Perhaps `>=14.0.0` or `14.x || 15.x`?

packages/ffe-file-upload-react/src/FileUpload.js
henrikhermansen: > File is an object, so even if we have duplicates, they should not be equal, right?

In the prop types comments here it says `files` is "A map of files, indexed by file name", meaning that `file` is not an object, but `files[file]` is.

However, it occurs to me now that of course the `files` object can't have duplicate keys, but how does this data structure work if the user wants to upload two files with identical file names? :thinking:

packages/ffe-file-upload/package.json
jafeld: Regarding the file structure, it's by design the files you upload should not contain duplicates, however, the logic enforcing this currently rest in the implementation, which is not ideal.

In the example, if you upload a file with same name, you overwrite the old one.

henrikhermansen: Okay. That is fine by me, just wanted to make sure we are aware of it, and that we've made a conscious decision that it's fine this way :)

henrikhermansen (dismissed):

henrikhermansen (approved):

PR #699, opened 2019-09-16
Add white as background color to gridCol and gridRow

lfbergee:
In some cases in KFP-applications we have a background color on the body or a wrapper div, and use the grid to contain the content. As it's now we have to color the content inside the grid and add margin to that to get the desired look, if we want a white background on the content. This is not a problem when adding a blue background to the grid, then we can use the color and margin from the grid.

With this change we can explisitly set white as background color, that will be over the body color and we can use the grid properly in the case of white grid over colored background.

### DIFF
--- a/packages/ffe-grid-react/src/GridCol.js
+++ b/packages/ffe-grid-react/src/GridCol.js
@@ -151,6 +151,7 @@ GridCol.propTypes = {
         'orange-salmon',
         'red',
         'blue-sky',
+        'white',
     ]),
     /** Any extra classes are attached to the root node, in addition to ffe-grid__col classes */
     className: string,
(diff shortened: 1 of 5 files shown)


ivarni (approved):
👍

ivarni:
Thanks! We do need this PR to be against the `develop` branch though. Then the bot that handles releases will later merge it to `master`. I don't think it's currently possible to manually merge anything to the `master` branch.

antidecaf (changes requested):
Thanks for the PR!

Like @ivarni says it needs to go towards `develop`, other than that it looks good. :)

lfbergee:
Thx, changed base to `develop`

antidecaf (approved):

lfbergee:
Any idea when this will go into prod?

antidecaf:
The next scheduled publish is at 16.00.

ivarni:
There's an automatic script that does the publishing, it should run daily so hopefully there'll be a release out by tomorrow.

PR #712, opened 2019-10-02
New icon for caravan added to ffe-icons

maylindamartinsen:
DIG-55545


packages/ffe-icons/icons/campingvogn-ikon.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="200px" height="200px" viewBox="0 0 119 82" version="1.1" xmlns="http://www.w3.org/2000/svg">
```
ivarni: Could you remove the `width` and `height` properties? They cause a bit of a struggle when trying to size the image with CSS. We also don't need `version` and `xmlns` properties (though I don't know if they cause issues) and I suspect the corresponding react-package won't like the `<?xml>` element :)

maylindamartinsen: crap removed!

ivarni (approved):

PR #717, opened 2019-10-09
feat(ffe-datepicker): Add full width style for datepicker and dateinput

torkjels:
Styling som gjør det mulig å ha `Datepicker` i fullbredde-modus.

image

### DIFF
--- a/packages/ffe-datepicker/less/dateinput.less
+++ b/packages/ffe-datepicker/less/dateinput.less
@@ -2,6 +2,10 @@
     position: relative;
     display: inline-block;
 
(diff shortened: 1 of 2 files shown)


ivarni:
I'm thinking this should just be the default.

A long time ago we agreed that the components should take all available space to make page layout easier, so if they don't that should be considered a bug. 

But maye releasing this as a new feature first and then doing a breaking change to make it the default behaviour would cause less friction?

ivarni (approved):

antidecaf (approved):

PR #725, opened 2019-10-14
Develop center headings

pethel:
Centering of headings

### DIFF
--- a/packages/ffe-core-react/src/typography/Heading.js
+++ b/packages/ffe-core-react/src/typography/Heading.js
@@ -13,6 +13,7 @@ const makeHeading = level => {
             noMargin,
             withBorder,
             lookLike,
+            textCenter,
             ...rest
         } = props;
 
@@ -28,6 +29,7 @@ const makeHeading = level => {
                     { [`${headingClass}--inline`]: inline },
                     { [`${headingClass}--no-margin`]: noMargin },
                     { [`${headingClass}--with-border`]: withBorder },
+                    { [`${headingClass}--text-center`]: textCenter },
                     className,
                 )}
             >
@@ -44,6 +46,7 @@ const makeHeading = level => {
         lookLike: oneOf([1, 2, 3, 4, 5, 6]),
         noMargin: bool,
         withBorder: bool,
+        textCenter: bool,
     };
 
     return Heading;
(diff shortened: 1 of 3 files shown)


antidecaf (approved):

packages/ffe-core-react/src/typography/Heading.js
```diff
@@ -28,6 +29,7 @@ const makeHeading = level => {
                     { [`${headingClass}--inline`]: inline },
                     { [`${headingClass}--no-margin`]: noMargin },
                     { [`${headingClass}--with-border`]: withBorder },
+                    { [`${headingClass}--text-center`]: textCenter },
```
ivarni, line 32: Can we please rename this to `centerText` like for [the grid](https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-grid-react/src/GridCol.js#L41) (or rename the grid modifier to `textCenter`, I don't really care as long as it's consistent).

packages/ffe-core-react/src/typography/Heading.js
```diff
@@ -44,6 +46,7 @@ const makeHeading = level => {
         lookLike: oneOf([1, 2, 3, 4, 5, 6]),
         noMargin: bool,
         withBorder: bool,
+        textCenter: bool,
```
hdaljord, line 49: Can you please add this one to index.d.ts as well?

packages/ffe-core-react/src/typography/Heading.js
pethel: I am now consistent with Paragraph that also uses "textCenter". I though that would make more sense because they are both text element?

packages/ffe-core-react/src/typography/Heading.js
pethel: Is it maybe a way to automate this? I think maintaining 2 set of "prop-types" is a bit unfortunate.

hdaljord: Maybe, but I don't have time to look into this now.

packages/ffe-core-react/src/typography/Heading.js
ivarni: If we're already inconsistent across modules I guess it doesn't really matter. Hopefully at some point we'll have a dedicated team that can also work to keep the various APIs predictable :)

packages/ffe-core-react/src/typography/Heading.js
pethel: https://github.com/SpareBank1/designsystem/pull/729

PR #728, opened 2019-10-15
fix(ffe-grid): Remove screen type from media queries, see #719

torkjels:
Removing the `screen` part of the media queries in `ffe-grid` as discussed in https://github.com/SpareBank1/designsystem/issues/719.

### DIFF
--- a/packages/ffe-grid/less/ffe-grid.less
+++ b/packages/ffe-grid/less/ffe-grid.less
@@ -151,7 +151,7 @@
         padding-top: 0;
     }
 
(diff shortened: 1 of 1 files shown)


ivarni:
If you could put that first comment into the commit message instead of "Fix grid print styles" that would be great, because that's what's going to end up in the changelog. 

It's going to be difficult for people who are checking the log before updating to guess what was changed based on the current commit message.

antidecaf (approved):

ivarni (approved):

ivarni:
I guess it's time to stress-test the release script :shipit:

PR #748, opened 2019-10-21
feat(ffe-form-react): InputGroup `hideErrorOnChange` set to `true` hi…

torkjels:
…des error message on change and renders it on blur

Setting the `hideErrorOnChange` prop to `true` hides the error message on change, and renders it on blur. This means that you won't see an error message until the `InputGroup` has lost focus.

Example:
With `hideErrorOnChange` set to false (default), writing an email address into an input field will display an error message while the user is typing, until the complete valid email is typed.
With `hideErrorOnChange` set to true, the error message will only display after the input group loses focus, and will be hidden again if the user goes back to the input field and starts to type.

### DIFF
--- a/packages/ffe-form-react/src/InputGroup.js
+++ b/packages/ffe-form-react/src/InputGroup.js
@@ -9,8 +9,21 @@ import ErrorFieldMessage from './ErrorFieldMessage';
 class InputGroup extends Component {
     constructor(props) {
         super();
+        this.onChange = this.onChange.bind(this);
+        this.onBlur = this.onBlur.bind(this);
+        this.state = {
+            isEditing: false,
+        };
 
         this.id = props.inputId ? props.inputId : `input-${uuid.v4()}`;
+    }
+
+    onChange() {
+        this.setState({ isEditing: true });
+    }
+
+    onBlur() {
+        this.setState({ isEditing: false });
     }
 
     render() {
@@ -25,8 +38,15 @@ class InputGroup extends Component {
             fieldMessage,
             tooltip,
             onTooltipToggle,
+            hideErrorOnChange,
             ...rest
         } = this.props;
+
+        const { isEditing } = this.state;
+
+        const shouldShowError = hideErrorOnChange
+            ? !isEditing && !!fieldMessage
+            : !!fieldMessage;
 
         if (React.Children.count(children) > 1) {
             throw new Error(
@@ -55,7 +75,7 @@ class InputGroup extends Component {
         const extraProps = {
             id: this.id,
             'aria-invalid': String(
-                !!fieldMessage &&
+                shouldShowError &&
                     (typeof fieldMessage === 'string' ||
                         fieldMessage.type === ErrorFieldMessage),
             ),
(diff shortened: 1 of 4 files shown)


ivarni:
Do we want to have different timing for when errors are shown in different applications? And if so, can't this be solved in user-land by making a wrapper-component that uses `onBlur` to achieve this if it's really needed so we can keep an API that encourages similar behaviour across applications and keep exceptions to the general flow isolated?

packages/ffe-form-react/src/InputGroup.js
```diff
@@ -66,14 +86,25 @@ class InputGroup extends Component {
...
             <div
                 className={classNames(
                     'ffe-input-group',
                     { 'ffe-input-group--no-extra-margin': !extraMargin },
-                    { 'ffe-input-group--error': !!fieldMessage },
+                    { 'ffe-input-group--error': shouldShowError },
                     className,
                 )}
+                onChange={this.onChange}
+                onBlur={this.onBlur}
```
torkjels, line 107: Because of this change, the linter throws an error saying "Static HTML elements with event handlers require a role  jsx-a11y/no-static-element-interactions".
Adding `onChange` and `onBlur` to a `div` generally feels a bit.. weird - but it works! But I don't know what aria role to give this div.

Should I
a) Give the div some role?
b) Ignore the linter rule
c) Pass the `onBlur` and `onChange` down in `extraProps` doing something along the lines of
```
const extraProps = {
  ......,
  onBlur: () => {
    this.onBlur();
    if(this.props.onBlur) {
      this.props.onBlur()
    }
  }
}
```
Solution c) might risk the `onBlur` and `onChange` never being called if they're never applied in the children.

torkjels:
> Do we want to have different timing for when errors are shown in different applications? And if so, can't this be solved in user-land by making a wrapper-component that uses `onBlur` to achieve this if it's really needed so we can keep an API that encourages similar behaviour across applications and keep exceptions to the general flow isolated?

That's a great point. In our application we've come to use an approach like this as the standard (wrapping it in a component with the onBlur logic like you mentioned) for every input field. But you're right in saying it should be similar behavior across the applications.

Do we know how different applications handle this? I wouldn't be surprised (a view based on absolutely nothing except for my own hunch) if several applications implement some sort of behavior like this.

It might be worth having a discussion about this with e.g. some designers :art: :thinking:

ivarni:
We're sort of already waiting for them to get back to us on it, ref #216 :)
(That's usually how a lot of these kinds of broader discussions end unfortunately)

From the viewpoint of `InputGroup` the way it works right now it just reflects the state that the application hands it. If it's told there's an error then it shows the error but it does not concern itself about *when* there's an error. It's primarily a visual component. I think that makes it easier to deal with and reason about and keeps the API more lean. I'd actually rather see the tooltip parts removed than more logic added to it, and especially now that it's based on the function-as-a-child pattern it would be more natural to let the consumer compose in both the tooltip and the error message and have both as separate components.

packages/ffe-form-react/src/InputGroup.js
```diff
@@ -127,10 +153,12 @@ InputGroup.propTypes = {
     onTooltipToggle: func,
     /** Use the Tooltip component if you need more flexibility in how the content is rendered. */
     tooltip: oneOfType([bool, node]),
+    hideErrorOnChange: bool,
```
hdaljord, line 156: Could you please add this new prop to index.d.ts as well?

torkjels: :+1: done

henrikhermansen:
I'm sorry, but I agree with @ivarni that this seems like an unnecessary complication to the API.
The `fieldMessage` prop does a good job of doubling as a switch for whether to show an error message or not.

I also want to question the UX part of this behavior.
If I use this input field and enter some invalid value, I would get the error on blur.
However, as soon as I start typing in the field again, the error would go away, because I am editing.
Personally, this would give me the impression that the change I just made was sufficient, and I would be annoyed to see the error reappear on blur again. Because of this it might require a lot of guesswork by the user to get the field value right. Is this behavior something UX would really condone?
Unless I misinterpreted the code changes in this PR?

To answer your question about what others are doing. In our applications we do not display any error messages until the user tries submitting the form. From then on we will update error messages on the fly when the user makes changes. And yes, we apply this logic to what we put into the `fieldMessage` prop.

PR #753, opened 2019-10-30
feat(ffe-core): export spacing to JS

henrikhermansen:
This makes some slight changes to the build script used to export less variables as JS variables.

In order to export the spacing variable from spacing.less we need to somehow handle variables pointing to other variables. For example `@ffe-spacing-2xs: (@ffe-spacing * 0.5); // 4px`.

My approach to this is simply to filter out any variables that points to other variables.

These changes does not cause any changes to the generated `breakpoints.js` or `colors.js`.

### DIFF
--- a/packages/ffe-core/bin/build.js
+++ b/packages/ffe-core/bin/build.js
@@ -4,7 +4,7 @@ const Case = require('case');
 const mkdirp = require('mkdirp');
 
 // These are the files we want to convert to JavaScript
-const FILES_WITH_VARIABLES = ['colors', 'breakpoints'];
+const FILES_WITH_VARIABLES = ['colors', 'breakpoints', 'spacing'];
 
 // First, create the lib directory if it doesn't exist
 mkdirp.sync(path.resolve(__dirname, '..', 'lib'));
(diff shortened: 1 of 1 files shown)


packages/ffe-core/bin/build.js
```diff
@@ -22,9 +22,10 @@ FILES_WITH_VARIABLES.forEach(filename => {
         .split('\n')
         .map(line => line.trim())
         .filter(line => line.startsWith('@') && !line.startsWith('@import'))
-        .map(line => line.replace('@ffe-', ''))
-        .map(line => line.replace('@', ''))
         .map(line => line.split(': '))
+        .filter(([, value]) => !value.includes('@'))
+        .map(([key, value]) => [key.replace('@ffe-', ''), value])
+        .map(([key, value]) => [key.replace('@', ''), value])
```
henrikhermansen, line 28: The change to these two lines was made because I moved the split on `': '` up, in order to make line 26 possible.
I figured the replace on `line` was done with the intent of "renaming" the variable, thus doing the replace on `key` will now be equivalent.

ivarni:
I've no idea how any of this works but I trust you've tested it :)

ivarni (approved):

PR #756, opened 2019-11-06
fix(ffe-form-react): allow number values in prop types

henrikhermansen:
This will make it easier to work with numerical values in forms, as you will only have to parse the value when it is read from the field (because fields always "returns" strings)
Prior to this change you would need to parse numbers to strings just to satisfy prop types, or settle for having numerical values as strings in your form state.


pethel (approved):
lgtm

ivarni (approved):

PR #767, opened 2019-11-22
docs(ffe-grid-react): Reword GridCol size validity warning.

torkjels:
Our team spent a long time to understand what the `GridCol` warning meant.
For example, we got this warning
```
The grid should have 4 columns for "sm" screens, please stick to a total of 4 columns.
For "sm" screensizes this means you need to stick to multiples of 3 for cols+offset.
Please consult the ffe-grid README.md for details.
You've provided a <GridCol> with sm={{"cols":4,"offset":0}}
```
It says we should stick to 4 columns, but as the warning itself says, we have supplied `cols: 4`, so from our point of view we **did** stick to 4 columns.

This edit makes it clear that the only thing that matters is that you stick to the correct multiples.

### DIFF
--- a/packages/ffe-grid-react/src/GridCol.spec.js
+++ b/packages/ffe-grid-react/src/GridCol.spec.js
@@ -301,7 +301,7 @@ describe('GridCol', () => {
 
                 expect(console.error).toHaveBeenCalledTimes(1);
                 expect(console.error.mock.calls[0][0]).toContain(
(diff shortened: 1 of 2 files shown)


ivarni:
Yeah, we were thinking of changing the API at som point to make it less math-y but never got around to it...
Not sure there's much point right now, it will have to undergo some major breaking changes at some point anyway.

ivarni (approved):

PR #771, opened 2019-11-25
fix(ffe-datepicker-react): fix broken datepicker for JS users

JonasJore:
Fix for import statement in index at root where import for the Datepicker component was instead referencing index at root level (itself).

Also changed interface name for `InputProps` in `index.d.ts` at root since prefixing name with I for interface is a bit redundant.

### DIFF
--- a/packages/ffe-datepicker-react/src/index.d.ts
+++ b/packages/ffe-datepicker-react/src/index.d.ts
@@ -1,6 +1,6 @@
 import React from 'react';
 
-interface IinputProps {
+interface InputProps {
     className?: string;
     id?: string;
 }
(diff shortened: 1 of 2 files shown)


packages/ffe-datepicker-react/src/index.d.ts
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 
-interface IinputProps {
```
hdaljord, line 3: Flott å endre fra IinputProps, men husk å endre også på linje 15 der denne brukes

hdaljord (changes requested):

packages/ffe-datepicker-react/src/index.d.ts
JonasJore: Fikset det nå

henrikhermansen (approved):

hdaljord (approved):

PR #779, opened 2019-11-29
chore: bump commitlint and add config-conventional

henrikhermansen:
I've noticed some misconceptions about conventional commits in various commit messages. While I've made some people aware of it, I can't catch them all, and I don't really want to be a conventional commit police anyways.

Time to add some more rules through the conventional config from commitlint :)


ivarni:
I thought we already had this, at least I've been stopped from commiting before, but more checks never hurts anyone 👮

ivarni (approved):

henrikhermansen:
> I thought we already had this, at least I've been stopped from commiting before, but more checks never hurts anyone

We just had enforcement on a type (any type, you could just make something up), now it's restricted to an enum of sensible types.
Also we had (and still have) enforcement on scope restricted to an enum we've defined. There's a package `@commitlint/config-lerna-scopes` which probably does some magic, but I didn't wanna mess with our scope enum so I didn't add it.

Now we have [a few more rules](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional), all of which I think makes sense.

ivarni:
I guess I never was enough of a rebel to try custom types :)

PR #783, opened 2019-12-04
Add radio step btn

Amanej:
Added a horizontal selector, used in the car loan flow. That also will be used in all object loan flows, consumer credit loan application, "utsett-betaling" applikasjoner etc.

This is a request from team hverdagsfinansiering

Screenshot from 2019-12-03 12-37-49


henrikhermansen (approved):
Commits and code looks good now.
Awaiting UX forum 18th of December to see if this component should be included in the Designsystem.

pethel (approved):
Code looks good now so I approve. This can be merged after UX has made their decision.

PR #785, opened 2019-12-06
Set default text styling to be non selectable 

Amanej:
This will set paragraphs and body text to be not selectable by default.
This an issue especially in the mobile app. When we use the base text components in a context where users are frequently pressing buttons and then experiences the unintended of text highlighting in other elements.

In the cases where one intends the user to be able to select and copy text, one should be explicit.

It is based on the [pending pull request](https://github.com/SpareBank1/designsystem/pull/784).

### DIFF
--- a/packages/ffe-core/less/typography.less
+++ b/packages/ffe-core/less/typography.less
@@ -231,13 +231,14 @@
     font-family: 'MuseoSans-500', arial, sans-serif;
     line-height: 24px;
     .ffe-fontsize-body-text;
+    .ffe-no-select;
 }
 
 .ffe-body-paragraph {
     margin-bottom: 1em;
     margin-top: 0;
     line-height: 24px;
-
+    .ffe-no-select;
     &--text-center {
         text-align: center;
     }
@@ -251,6 +252,7 @@
 .ffe-sub-lead-paragraph {
     font-family: 'MuseoSans-500', arial, sans-serif;
     margin-top: 0;
+    .ffe-no-select;
 }
 
 .ffe-lead-paragraph {
--- a/packages/ffe-core/less/typography.less
+++ b/packages/ffe-core/less/typography.less
@@ -318,6 +318,16 @@
     margin: 0 4px;
 }
 
+.ffe-no-select {
+    user-select: none;
+    -webkit-touch-callout: none;
+}
+
+.ffe-select {
+    user-select: auto;
+    -webkit-touch-callout: default;
+}
+
 @media (min-width: @breakpoint-sm) {
     .ffe-h1 {
         line-height: 56px;
--- a/packages/ffe-core-react/src/typography/Paragraph.md
+++ b/packages/ffe-core-react/src/typography/Paragraph.md
@@ -7,6 +7,15 @@
 </Paragraph>
 ```
 
+**_Tekst som kan velges_**
+
+```js
+<Paragraph>
+    Her kan du velge tekst, f.eks med kontonr{' '}
+    <span className="ffe-body-text ffe-select">123456789101</span>
+</Paragraph>
+```
+
 **Lead paragraph**
 
 ```js


antidecaf (changes requested):
While this may be useful in the context of web views within the mobile app, it is probably not a good idea for default styling.

packages/ffe-core/less/typography.less
```diff
@@ -318,6 +318,16 @@
     margin: 0 4px;
 }
 
+.ffe-no-select {
+    user-select: none;
+    -webkit-touch-callout: none;
```
pethel, line 325: Enig med @antidecaf  att dette ikke er lurt men  
dette fikser autoprefix. :)

Amanej:
> While this may be useful in the context of web views within the mobile app, it is probably not a good idea for default styling.

Why ? 
Because one should be able to copy any text on desktop, by default ? 
Or because there might be unintended consequences ? I can see that.
I cannot see why on mobile, one should not expect developers/designers to be explicit about text can be selected and copied.

But in that case, would be better if it only occurred in a web view context ? That is the main intent. Actually in general on mobile.

antidecaf:
> Why ?
> Because one should be able to copy any text on desktop, by default ?

Yes. This is a standard feature of all browsers. I don't see why we would disable it to begin with?

> I cannot see why on mobile, one should not expect developers/designers to be explicit about text can be selected and copied.

The ability to select and copy text is still the default behaviour in mobile browsers. If developers/designers want to disable it, then that explicit choice should probably be made on a case by case basis.

> But in that case, would be better if it only occurred in a web view context ?

Yes, that is a much better approach.

Edit: Closing this one for now. Going forward, let's discuss less obtrusive ways to fix this issue for web views.

Amanej:
> > Why ?
> > Because one should be able to copy any text on desktop, by default ?
> 
> Yes. This is a standard feature of all browsers. I don't see why we would disable it to begin with?
> 

I will give you the exact scenario I experienced to understand why this was suggested.
On mobile, you are pressing buttons to complete actions. Even in a browser.

Suddenly, you have selected a bunch of text you did not intend to.

Secondly, most users are not even in a browser, or at least do not believe they are in a browser.

For both of these users, mobile users, select text should not be a default behavior. Just try to use the app/website on mobile with empathy and you will experience it.

> > I cannot see why on mobile, one should not expect developers/designers to be explicit about text can be selected and copied.
> 
> The ability to select and copy text is still the default behaviour in mobile browsers. If developers/designers want to disable it, then that explicit choice should probably be made on a case by case basis.
> 

On desktop, I agree you are quite right.

But not in touch based interfaces, that is what was pointed out in the original.

> > But in that case, would be better if it only occurred in a web view context ?
> 
> Yes, that is a much better approach.
> 
> Edit: Closing this one for now. Going forward, let's discuss less obtrusive ways to fix this issue for web views.

But why not let us continue working on it, instead of closing it ?

I agree, that focusing on mobile users and especially the webview context is correct. But by just closing much of the previous discussion/context is lost.

packages/ffe-core/less/typography.less
Amanej: Enig om autoprefix, og desktop. Ikke mobil.

Amanej:
> > > But in that case, would be better if it only occurred in a web view context ?
> > 
> > 
> > Yes, that is a much better approach.
> > Edit: Closing this one for now. Going forward, let's discuss less obtrusive ways to fix this issue for web views.
> 
> But why not let us continue working on it, instead of closing it ?
> 
> I agree, that focusing on mobile users and especially the webview context is correct. But by just closing much of the previous discussion/context is lost.

???
Still cannot see why this was closed so quickly.
I found a bug, and I am using **my free time** attempting to contribute.

_Is there an established way to detect if the environment is a webview app ?_

Otherwise, I will soon push a fix  where I try to add this class only in a webview context.

antidecaf:
No variation of `user-select: none` _by default_ is sustainable, afaic. The impact on the general user experience is just too big to merge something like this into components that are used by a variety of apps without warning.

A more feasible fix might be to target buttons within web views specifically. Whether or not a web view is being rendered is determined within each webapp.

PR #789, opened 2019-12-09
fix(stylelint-config-ffe): increase nesting depth

antidecaf:
During the process of adding dark mode overrides, the need to increase the `max-nesting-depth` rule has surfaced, as every override is a nested rule. This prevents the builds from failing and removes the need to disable stylelint rules in individual less files.

### DIFF
--- a/linting/stylelint-config-ffe/index.js
+++ b/linting/stylelint-config-ffe/index.js
@@ -18,7 +18,12 @@ module.exports = {
         ],
         'no-missing-end-of-source-newline': null,
         'number-leading-zero': null,
-        'max-nesting-depth': 3,
+        'max-nesting-depth': [
+            4,
+            {
+                ignoreAtRules: ['media'],
+            },
+        ],
         'media-feature-name-no-vendor-prefix': true,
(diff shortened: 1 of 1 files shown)


mvmagnussen (dismissed):

linting/stylelint-config-ffe/index.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
         ],
         'no-missing-end-of-source-newline': null,
         'number-leading-zero': null,
-        'max-nesting-depth': 3,
+        'max-nesting-depth': 5,
```
kwltrs: With `ignoreAtRules: ['media']` a `max-nesting-depth` of 4 should be enough. I.e.:

```javascript
'max-nesting-depth': [4, { ignoreAtRules: ['media'] }],
```

antidecaf: Fixed in 03e9c4c

kwltrs (approved):

pethel (approved):

mvmagnussen (approved):

PR #790, opened 2019-12-10
Darkmode

mvmagnussen:
Added inline dark mode styling overrides for all ffe packages. Removed old darkmode overrides file.


ivarni:
So what is the goal here, could the commit messages maybe be somewhat more descriptive and give someone who hasn't been directly involved a clue about what the change accomplishes?

Also, if all the darkmode things can be nested under a `.native` class, can we use postcss to make two builds, one where those classes are stripped and one where they're not?

Finally, can we please namespace the `.native` class so we don't accidentally mess with non-FFE styles? CSS is global.

antidecaf:
> Finally, can we please namespace the .native class so we don't accidentally mess with non-FFE styles? CSS is global.

I believe the reason `.native` is used is it already exists on `body` in web views rendered in the mobile app. Although I do agree a namespaced class would be better.

kwltrs:
I have to agree that .native is not a very good choice for a class name.

ivarni:
Wait, isn't this going to automatically apply dark styling to any application that is embedded in a mobile app view? That sounds like a profoundly bad idea to me, we have absolutely no guarantee how that will end up looking when custom/local styling is thrown into the mix.

mvmagnussen:
> > Finally, can we please namespace the .native class so we don't accidentally mess with non-FFE styles? CSS is global.
> 
> I believe the reason `.native` is used is it already exists on `body` in web views rendered in the mobile app. Although I do agree a namespaced class would be better.

Correct.

antidecaf:
> Wait, isn't this going to automatically apply dark styling to any application that is embedded in a mobile app view?

I'm not really sure I understand the question, that's pretty much what this does? `.native` is used to scope the overrides to mobile app views, as there isn't really a unified plan for dark mode in place for the web UI as of right now, afaik.

ivarni:
> I'm not really sure I understand the question, that's pretty much what this does? `.native` is used to scope the overrides to mobile app views, as there isn't really a unified plan for dark mode in place for the web UI as of right now, afaik.

I'm asking because we have a web application that is embedded in a web view (mobilbanken) and this looks like it will affect us directly while we've had no chance to test what the result will actually look like, nor do we have the resources to do so right now. A push for us to start supporting dark mode should come via the people managing our priorities and not as a minor update to the designsystem IMO.

> I believe the reason .native is used is it already exists on body in web views rendered in the mobile app

So if this breaks our UI when our app is embedded as a web view, and I'm fairly sure it will, we'd have to either delegate resources we don't have to fixing any visual issues or start sniffing for and removing the `.native` class you describe. Both sound like a very unpleasant scenario.

mvmagnussen:
> > I'm not really sure I understand the question, that's pretty much what this does? `.native` is used to scope the overrides to mobile app views, as there isn't really a unified plan for dark mode in place for the web UI as of right now, afaik.
> 
> I'm asking because we have a web application that is embedded in a web view (mobilbanken) and this looks like it will affect us directly while we've had no chance to test what the result will actually look like, nor do we have the resources to do so right now. A push for us to start supporting dark mode should come via the people managing our priorities and not as a minor update to the designsystem IMO.
> 
> > I believe the reason .native is used is it already exists on body in web views rendered in the mobile app
> 
> So if this breaks our UI when our app is embedded as a web view, and I'm fairly sure it will, we'd have to either delegate resources we don't have to fixing any visual issues or start sniffing for and removing the `.native` class you describe. Both sound like a very unpleasant scenario.

Don't add the native class to the body tag in your webapp until you and your team are dark mode compatible.

ivarni:
So it's not already added like the previous comment suggested?

palok:
> I'm asking because we have a web application that is embedded in a web view (mobilbanken) and this looks like it will affect us directly while we've had no chance to test what the result will actually look like, nor do we have the resources to do so right now. A push for us to start supporting dark mode should come via the people managing our priorities and not as a minor update to the designsystem IMO.

Dark mode is already in production. This PR is about a re-design of what already is in place. No new features are really added here. 
 
> So if this breaks our UI when our app is embedded as a web view, and I'm fairly sure it will, we'd have to either delegate resources we don't have to fixing any visual issues or start sniffing for and removing the `.native` class you describe. Both sound like a very unpleasant scenario.

As explained, the dark mode styles in ffe will only be applied if the body element has the "native" class. For NPA/Banking apps, this is added by the web template framework. If your team does not use this template framework, your content will not be affected. It will however sort of ruin the "dark mode experience" since it will be displayed as dark text on white background and not the other way around.

**Edit**: who puts the "close" button next to the submitbutton :rofl:

mvmagnussen:
> So it's not already added like the previous comment suggested?

Not in your app, just everyone else's.

ivarni:
This is the kind of information that should be included in the commit message so it makes it to the changelog, or at least be in the description of an issue linked to the PR.

And yeah, the close-button placement is terrible.

mvmagnussen:
> This is the kind of information that should be included in the commit message so it makes it to the changelog, or at least be in the description of an issue linked to the PR.
> 
> And yeah, the close-button placement is terrible.

As @palok says above, **this PR** does not introduce anything significantly new, apart from a refactor of code that is already in production.

packages/ffe-buttons/less/base-button.less
```diff
@@ -164,10 +269,17 @@
     border: solid 2px @ffe-grey-silver;
     box-shadow: 0 1px 2px 1px rgba(0, 0, 0, 0.05);
     color: @ffe-blue-azure;
-
-    &.ffe-button--dark {
```
palok, line 168: Vi trenger ikke denne lenger? Darkmode > dark theme?

palok (dismissed):

packages/ffe-buttons/less/base-button.less
mvmagnussen: Bankinterne applikasjoner hos forsikring kan bruke en dark-prop for å legge på sin egen, mørkere stil på et utvalg komponenter. Har ingenting med dark mode å gjøre, så bare å overse. Men de arver dessverre utilsiktet fra sine dark mode moder-komponenter, og derfor må jeg skrive noen ekstra linjer for at de skal se "normale" ut i styleguidist i dark mode.

antidecaf (approved):

palok (approved):

PR #791, opened 2019-12-11
fix(ffe-buttons-react): ensuring typedefinition for the to prop for B…

JonasJore:
The `to` prop belonging to `BaseButtonProps`, making a button function as a link missed typedefinitions for use in TypeScript code was missing. Therefore this is added in this PR.

### DIFF
--- a/packages/ffe-buttons-react/src/index.d.ts
+++ b/packages/ffe-buttons-react/src/index.d.ts
@@ -4,6 +4,7 @@ export interface MinimalBaseButtonProps extends React.HTMLProps<HTMLElement> {
     className?: string;
     element?: HTMLElement | string | React.ElementType;
     innerRef?: React.RefObject<HTMLElement>;
+    to?: string; //used in order to make buttons work with react-router functionality in typescript-files.
 }
 
 export interface BaseButtonProps extends MinimalBaseButtonProps {


kwltrs:
I don't know this `to`-attribute. How does it make a button to link?

kwltrs:
Is this related to react-router?

JonasJore:
Yeah, the `to` prop receives a `LocationDescriptor<S>` and basically tells what the destination of the link is i think. I understand it that the `element` prop tells the button it is a link which is extended from `MinimalBaseButtonProps`.

I think it may be better for `to` to also be included in `MinimalBaseButtonProps`

kwltrs:
Is there a better way to solve this? To me it seems wrong to include a definition of a prop belonging to a third-party framework (i.e. react-router).

ivarni:
I don't know typescript very well but I don't think you can express *if element is of type Link, there can be a to prop* in a definition file.

I don't think it hurts to add an optional prop to solve this, but maybe throw in a comment to explain why?

JonasJore:
@ivarni would i comment this in the code or in the PR?

ivarni:
@JonasJore I'm a fan of doing it in the code, if `.ts` files support comments. PRs are a bit ephemeral.

ivarni (dismissed):
Would be better if `react-router` was mentioned specifically as there's nothing generic about the `to` prop but mentioning why it's there is still better than not :)

ivarni (approved):

PR #796, opened 2019-12-18
Common spacing in components

antidecaf:
This PR replaces hardcoded spacing values with universal variables defined in [`ffe-core`](https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-core/less/spacing.less). The values are based on the [8pt grid](https://builttoadapt.io/intro-to-the-8-point-grid-system-d2573cde8632) spatial system.

For the most part, hardcoded values could be replaced by a more or less equivalent variable. Some values could have potentially been replaced by several different variables (e.g. do I replace 20px with 16px or 24px?). I've tried to pick the ones that make the most sense on a case by case basis.

This _will_ cause visual diffs all over the place. :see_no_evil:

### DIFF
--- a/packages/ffe-accordion/less/ffe-accordion.less
+++ b/packages/ffe-accordion/less/ffe-accordion.less
@@ -82,7 +82,7 @@
 
     .ffe-accordion--blue & {
         border-top: 0;
-        margin-bottom: 10px;
+        margin-bottom: @ffe-spacing-xs;
 
         &:last-of-type {
             border-bottom: 0;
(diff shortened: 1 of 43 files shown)


ivarni:
Great to see this moving forward!

I'm wondering if peerDeps around the board can just be changed from `x` to `> x`  just to pre-empt the inevitable chaos releasing a major of every package usually unleashes? :)

ivarni (dismissed):

pethel (dismissed):

kwltrs (approved):

antidecaf:
@ivarni Sounds like a good idea to me, but I'm not sure how to work that out with Lerna.

ivarni:
Fair point, nothing seems to work out very well with Lerna :)

PR #821, opened 2020-01-08
fix(ffe-form): Changing margin in input-group from initial to auto

etse:
Initial does not work for IE11, while auto does. This introduced problems in IE where the 32px margin would not be overridden to initial (this creates a unwanted 32px margin in account-selector between dropdown and selected account-number)

Example of bug:
bilde


ivarni (approved):
`auto` is not the same as `initial` but as long as it looks OK in all browsers I say go for it.

PR #831, opened 2020-01-16
feat(ffe-form-react): added support for passing ref to radio buttons

mortenk:
This version offers new optional properties (innerRef on RadioButton and RadioBlock, leftInnerRef and rightInnerRef on RadioSwitch), which is passed as refs to input elements
On BaseRadioButton, selectedValue updates the defaultChecked attribute instead of checked attribute, so that re-render is no longer required when radio buttons are clicked

### DIFF
--- a/packages/ffe-form-react/src/BaseRadioButton.js
+++ b/packages/ffe-form-react/src/BaseRadioButton.js
@@ -7,6 +7,8 @@ import {
     oneOfType,
     shape,
     string,
+    func,
+    object,
     number,
 } from 'prop-types';
 import uuid from 'uuid';
@@ -27,6 +29,7 @@ class BaseRadioButton extends Component {
             tooltip,
             tooltipProps,
             value,
+            innerRef,
             dark,
             ...inputProps
         } = this.props;
@@ -38,7 +41,7 @@ class BaseRadioButton extends Component {
             className,
         );
 
-        const isSelected = checked || selectedValue === value;
+        const isSelected = selectedValue ? selectedValue === value : checked;
 
         return (
             <Fragment>
@@ -48,6 +51,7 @@ class BaseRadioButton extends Component {
                         'ffe-radio-input--dark': dark,
                     })}
                     id={this.id}
+                    ref={innerRef}
                     type="radio"
                     checked={isSelected}
                     value={value}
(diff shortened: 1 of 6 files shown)


packages/ffe-form-react/src/BaseRadioButton.js
```diff
@@ -48,8 +50,9 @@ class BaseRadioButton extends Component {
                         'ffe-radio-input--dark': dark,
                     })}
                     id={this.id}
+                    ref={innerRef}
                     type="radio"
-                    checked={isSelected}
+                    defaultChecked={isSelected}
```
ivarni: This feels very wrong. The component will re-render when its props change, wether `checked` or `defaultChecked` is used on a child should have no effect on that?

ivarni: I pasted this into the Checkbox.md locally and used `npm start` to run the designsystem docs

```
initialState = { checked: true };
<>
    <fieldset className="ffe-fieldset">
        <legend className="ffe-form-label">Hva er du interessert i?</legend>
        <Checkbox name="interests" value="sport" inline={false} defaultChecked={state.checked}>
            Sport
        </Checkbox>
        <Checkbox name="interests" value="food" inline={false}>
            Matlaging
        </Checkbox>
        <Checkbox name="interests" value="moon" inline={false}>
            Månen
        </Checkbox>
    </fieldset>
    <button onClick={() => setState({ checked: !state.checked})}>swap</button>
</>
```

Using `defaultChecked`, the checkbox does not update when the state changes. Using `checked` it does. So `defaultChecked` is only used during the initial render like the documentation says. If it works in your usecase you're most likely remounting the component on every change. You can check that using `useEffect` or `componentDidMount`.

mortenk: I can see how it would break things. I've tried a different approach, that leaves the checked attribute undefined when selectedValue is undefined, leaving state management to the browser.

packages/ffe-form-react/src/BaseRadioButton.js
```diff
@@ -38,7 +41,7 @@ class BaseRadioButton extends Component {
             className,
         );
 
-        const isSelected = checked || selectedValue === value;
+        const isSelected = selectedValue ? selectedValue === value : checked;
```
ivarni, line 44: Since `selectedValue` will now override `checked` if both are passed as props instead of the other way around, this should probably be a breaking change. Adding **BREAKING CHANGE** to the commit-message along with a description of what's changed will take care of that. 

It might cause regressions if anyone was relying on the previous behaviour so it's better to be explicit about it.

mortenk: Added "BREAKING CHANGE" to the commit message

ivarni: Then I don't have any further objections :)

ivarni (approved):

PR #835, opened 2020-01-21
fix: darkmode for iOS

palok:
Fixed an issue related to manual switching to darkmode
in iOS/MacOS

### DIFF
--- a/src/assets/darkmode-switch.js
+++ b/src/assets/darkmode-switch.js
@@ -19,7 +19,11 @@ function swapMediaQueryConditionText(fromText, toText) {
         rules = sheet.cssRules || sheet.rules;
         for (ruleIndex = 0; ruleIndex < rules.length; ++ruleIndex) {
             rule = rules[ruleIndex];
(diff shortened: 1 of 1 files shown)


mvmagnussen (dismissed):

src/assets/darkmode-switch.js
```diff
@@ -57,6 +61,13 @@ let animateCssTransition = () => {
     }, 1000);
 };
 
+function isSafari() {
+    var ua = navigator.userAgent.toLowerCase();
+    if (ua.indexOf('safari') != -1 && ua.indexOf('chrome') != -1) {
+        return true;
```
JonasJore, line 67: Wont this function return `undefined` if the if-condition is false?

palok: You're right. It works now because `undefined` is falsy. But I will add an else clause to clarify

JonasJore (dismissed):

mvmagnussen (approved):

PR #836, opened 2020-01-23
fix(ffe-grid): larger GridCol padding on small devices

tutturen:
We have an issue with GridCol having a horizontal padding of only `8px` on small devices. This makes content closer to the edge of the screen than we like it to be, and escpecially close for devices with curved glass.

The proposed solution is to increase horizontal padding on `GridCol` for small devices from `8px` to `16px`.

### DIFF
--- a/packages/ffe-grid/less/ffe-grid.less
+++ b/packages/ffe-grid/less/ffe-grid.less
@@ -300,7 +300,7 @@
     flex: 1 1;
     flex-direction: column;
     max-width: 100%;
-    padding: 0 (@ffe-grid-gutter-sm / 2) @ffe-grid-gutter-sm;
+    padding: 0 @ffe-grid-gutter-sm @ffe-grid-gutter-sm;
 
     @media (min-width: @breakpoint-lg) {
         padding: 0 (@ffe-grid-gutter-lg / 2) @ffe-grid-gutter-lg;


ivarni:
This would increase the padding between adjacent columns as well, leaving us even less horizontal space for content. If the issue is with the edges of the screen then we should instead add outer padding to the entire grid.

https://github.com/SpareBank1/designsystem/blob/e8a03e6bf335260515a72c0f184aae37521b69c4/packages/ffe-grid/less/ffe-grid.less#L147

That means someone running the full 4 col width won't lose an additional 24px of horizontal space.

Either way, this might break existing layout and should be a major version, not a patch.

ivarni:
@antidecaf Has there been any discussion on this in regards to the space project? The old padding was 10px so the change to 8px was a 20% reduction, was that intentional?

tutturen:
Agreed @ivarni. This has an effect on more than just margins on the edge of the screen. We often use background color on GridCol elements, which also span the entire screen on small devices.

Could having a space on the entire grid making full width background harder?

henrikhermansen:
> @antidecaf Has there been any discussion on this in regards to the space project? The old padding was 10px so the change to 8px was a 20% reduction, was that intentional?

Just wanna add that even 10px was a bit narrow on mobile devices at the screen edges.

henrikhermansen:
> If the issue is with the edges of the screen then we should instead add outer padding to the entire grid.

Like @tutturen points out, that might be problematic with grid col background colors if you want them to fill the screen width.
What about some first-child and last-child magic to add padding to the first and last col in a row?

ivarni:
I think that when the `background` modifier is set on a grid-row it spans the entire width of the grid so yes I think padding could create an issue for that.

Maybe first/last selectors can be used to only add extra padding to the "outer" column(s)?

henrikhermansen:
> I think that when the `background` modifier is set on a grid-row it spans the entire width of the grid so yes I think padding could create an issue for that.
> 
> Maybe first/last selectors can be used to only add extra padding to the "outer" column(s)?

For some reason the grid row has flex-wrap:wrap, so cols might be stacked vertically. That wouldn't work too well with first and last child selectors.

I thought backgrounds were set on cols, but I see now you can set them on rows or cols, as you like. So I guess we could add the extra padding on the rows instead of the entire grid?

ivarni:
> I thought backgrounds were set on cols, but I see now you can set them on rows or cols, as you like. So I guess we could add the extra padding on the rows instead of the entire grid?

I thought we only supported backgrounds on rows but nevertheless I suppose padding the row might work. We do change the DOM-structure a little for a non-background row vs a background-row so both variants should be tested.

antidecaf:
> @antidecaf Has there been any discussion on this in regards to the space project? The old padding was 10px so the change to 8px was a 20% reduction, was that intentional?

The change was intentional indeed. All spacing values are supposed to adhere to the 8px grid, and any hardcoded instance of 10px has generally been translated into 8px since that is the closest value among our variables. Not sure if 8px or 16px is the answer in this particular case though, I suppose that's a question for designers. (@sirioc? @Heilysto?)

tutturen:
I have had a chat with Anne Granhaug (designer) which in turn have spoken to @sirioc and @Heilysto. The conclusion right now is to increase the gutter spacing on the sides to 16px.

On a more general note, she wanted to streamline the spacing in our design accross all pages, because right now there are no specific rules for spacing?

henrikhermansen:
Sounds like you want to put that side padding on the grid row then, to avoid increasing gutter size between cols.

ivarni:
And if we're doing this, please make it a breaking change. There's a high chance of breaking existing layouts, at least in our apps and probably others as well.

steffenz:
After leaving this open for a while, we've decided that the grid needs a complete overhaul to a more modern CSS based version. With this is mind, it seems most reasonable to not introduce any new changes before finally replacing/and or deprecating it. 
If you have any feedback or feel like this is a mistake, please let us know!

PR #837, opened 2020-01-23
fix(ffe-form-react): correctly handle non-undefined selected value

henrikhermansen:
When a selected value is provided it should be used to compare it to the
radio button value. Any value except undefined should be considered as a
provided value, even if it is falsy.

In my code my selected value is `''` (empty string) when no button is selected, and I don't provide any value for `checked`. This caused `isSelected` to become `undefined`, and React treats the input as an uncontrolled component.
When a radio button was selected, I got a warning about uncontrolled->controlled component in the console.

### DIFF
--- a/packages/ffe-form-react/src/BaseRadioButton.js
+++ b/packages/ffe-form-react/src/BaseRadioButton.js
@@ -41,7 +41,8 @@ class BaseRadioButton extends Component {
             className,
         );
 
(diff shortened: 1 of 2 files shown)


mortenk (approved):
Ser greit ut for meg

ivarni (approved):

PR #848, opened 2020-02-10
Generate typescript variable declarations

tutturen:
## Problem 1: ffe-core problematic .ts-files 
I recently merged a change to `ffe-core` which implemented generation of typescript declarations for constants defined in `.less`-files. This meant that `.ts`-files were published (including `.js` and `.d.ts` files), which in turn lead to confusion for the typescript compiler and lots of error messages.

### solution
The solution was to only publish `.js` and `d.ts` files, so this PR deletes the `.ts` files after they are used for type declarations.

## Problem 2: generate typescript declaration files for all Icons in  ffe-icons-react
@henrikhermansen suggested that we also add typescript types all the `Icon` components in `ffe-icons-react`.

### solution
My suggested solution is to replace `babel` with `tsc` in order to use that to generate declaration files from the `.tsx` files generated from the build-script. This could have unintented consequenses as the final transpiled code will be a bit different, but the behaviour should be the same. Another solution I thought about would be to just generate the `.d.ts` files directly with a template and keep the old `babel` setup, but I thought that could lead to problems down the line should the component change.

### DIFF
--- a/packages/ffe-core/bin/build.js
+++ b/packages/ffe-core/bin/build.js
@@ -6,7 +6,8 @@ const mkdirp = require('mkdirp');
 // These are the files we want to convert to JavaScript
 const FILES_WITH_VARIABLES = ['colors', 'breakpoints', 'spacing'];
 
-// First, create the lib directory if it doesn't exist
+// First, create the lib and tmp directories if they don't exist
+mkdirp.sync(path.resolve(__dirname, '..', 'tmp'));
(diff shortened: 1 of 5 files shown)


Snikanes:
Looks good to me! A renaming of `build-jsx-components.js` to `build-tsx-components.js` might be in order, though!

Snikanes (dismissed):

Snikanes (approved):

JonasJore (approved):
looks good.

PR #862, opened 2020-03-02
Extend SVG attributes for icons in TypeScript Interface

### LINKED ISSUE #860: ffe-icons-react: IconProps interface does not allow for rest parameters
## Which package is this issue related to?
ffe-icons-react

After upgrading ffe-icons-react to 7.2.2 (from 7.0.3, typescript declarations for all icons was added in the 7.1.0) I get a type errors like this

`ERROR in /home/sb1a.sparebank1.no/a500118/git/kundeklient-forsikring-bm/kfb-frontend/src/main/components/Huskeliste/Oppgave.tsx(25,30):
TS2322: Type '{ className: string; }' is not assignable to type 'IntrinsicAttributes & IconProps & { children?: ReactNode; }'.
  Property 'className' does not exist on type 'IntrinsicAttributes & IconProps & { children?: ReactNode; }'.
`

The code:
`
<LyspareIkon className="kfb-icon--sm kfb-icon--blue-royal"/>
`

As the generated component accepts rest-parameters standard HTML attributes, such as className, should be allowed.

IconProps should probably extend React.HTMLAttributes<HTMLElement> .

steffenz:
## Issue description
When using icons with TypeScript, attributes not defined in the IconProps triggers an error.
Allowed SVG attributes should be supported as props to these elements. 
Related issue: https://github.com/SpareBank1/designsystem/issues/860 

## Solution
The short and simple solution here is extending React.SVGAttributes and spread them in with the other props defined. ~~Please do note however that overriding props set in the embedded SVG icon itself will simply be ignored (for example ViewBox). If this ends up being an problem it should be reported in as a separate issue, as it is unrelated to adding type support.~~ Overriding SVG props is now supported.

### DIFF
--- a/packages/ffe-icons-react/src/build-tsx-components.js
+++ b/packages/ffe-icons-react/src/build-tsx-components.js
@@ -59,19 +59,17 @@ const svg = ${toTsx(icons[iconName])};
 
 interface IconProps extends React.SVGAttributes<SVGElement> {
     desc?: string;
-    focusable? : boolean;
     title?: string;
     iconName?: string;
 }
 
 const Icon: React.FC<IconProps> = ({
     desc,
-    focusable = false,
     title,
     iconName,
     ...rest
     }) => (
-        <svg focusable={String(focusable)} {...rest} {...svg.props}>
+        <svg {...svg.props} {...rest}>
             {title &&
                 <title>{title}</title>
             }
--- a/packages/ffe-icons-react/src/build-tsx-components.js
+++ b/packages/ffe-icons-react/src/build-tsx-components.js
@@ -53,13 +53,15 @@ const toTsx = svgString => {
  * */
 const createStandaloneTSX = (icons, iconName) => `
 import * as React from 'react';
-import { bool, string } from 'prop-types';
+import { string, bool } from 'prop-types';
+import * as PropTypes from 'prop-types';
 
 const svg = ${toTsx(icons[iconName])};
 
-interface IconProps extends React.SVGAttributes<SVGElement> {
+interface IconProps extends Omit<React.SVGAttributes<SVGElement>, 'focusable'> {
     desc?: string;
     title?: string;
+    focusable?: boolean | string;
     iconName?: string;
 }
 
@@ -82,8 +84,11 @@ const Icon: React.FC<IconProps> = ({
 
 Icon.propTypes = {
     desc: string,
-    focusable: bool,
     title: string,
+    focusable: PropTypes.oneOfType([
+        PropTypes.bool,
+        PropTypes.oneOf(["true", "false", "auto", "undefined"]),
+    ]),
     iconName: string,
 };
 
--- a/packages/ffe-icons-react/src/build-tsx-components.js
+++ b/packages/ffe-icons-react/src/build-tsx-components.js
@@ -57,7 +57,7 @@ import { bool, string } from 'prop-types';
 
 const svg = ${toTsx(icons[iconName])};
 
-interface IconProps {
+interface IconProps extends React.SVGAttributes<SVGElement> {
     desc?: string;
     focusable? : boolean;
     title?: string;


packages/ffe-icons-react/src/build-tsx-components.js
```diff
@@ -59,19 +59,17 @@ const svg = ${toTsx(icons[iconName])};
 
 interface IconProps extends React.SVGAttributes<SVGElement> {
     desc?: string;
-    focusable? : boolean;
```
JonasJore, line 62: @steffenz  So, `focusable` is a duplicate because it is inherited from `React.SVGAttributes`?

steffenz: That's right! I probably should have updated the description for that. This doesn't seem to break anything, and the new inherited "focusable" prop will accept both boolean and string enum values (allowing values such as as "true", "false" and "auto"). I didn't remove it from the PropTypes however to make this available to non-typescript users and the styleguidist documentation :smile:

hdaljord: But the propType definition of focusable is bool, as both ts-definition and propType should match I believe you should either add string to propType, or weaken the ts-definition.

JonasJore: I see, then this looks good to me :smile:

JonasJore (dismissed):

packages/ffe-icons-react/src/build-tsx-components.js
steffenz: I'm not sure how to solve this, @hdaljord . I've tried multiple ways of allowing both a boolean, boolish values in PropTypes - but the build keeps failing stating they are incompatible.
I tried various version of the one below, without any luck. 
```jsx
 focusable: PropTypes.oneOfType([
        PropTypes.bool,
        PropTypes.oneOf(["true", "false", "auto", "undefined"]),
    ])
```
When building I get the following error (in various variations):
```bash
  ../../node_modules/@types/react/index.d.ts:2381:9
    2381         focusable?: Booleanish | "auto";
                 ~~~~~~~~~
    The expected type comes from property 'focusable' which is declared here on type 'WeakValidationMap<IconProps>'
```
Essentially, just using ``bool`` will only allow true/false and not some of the other values (auto, undefined) which wasn't supported previously either. Not sure if we should bother fixing this?

hdaljord: I assume you tried using
`
focusable: oneOfType([bool, string])
`
as used in Datepicker?

Wouldn't you get an error from propTypes if you pass in a string when the propType definition is bool? Or am I missing something here?

steffenz: I've tried your suggested solution previously, but it does unfortunately lead to the same error. A mismatch between TypeScript interface and PropTypes. 

One possible solution is to omit "focusable" from the SVGElement (and thus overriding it) and then add the enum check in PropTypes. This will as far as TypeScript is concerned allow any string, but PropTypes will do a enum check and throw a console.error if it's not in the allowed list. I've updated the code with this change. What do you think about this way of solving it? :thinking:

ivarni: I can't comment on the TS bit but I don't think we absolutely need a 1:1 relationship between TS and propTypes. TS users will still see the propType warnings at runtime if I'm not mistaken, so it's not a big issue if TS is loser than propTypes. If that makes sense.

hdaljord: @steffenz, I think that is a better solution. And it even adds the possibility to pass in the valid values with propTypes. :+1: 

I think it's a good principle to have looser ts types and leave it to propTypes when it's not possible to make it the same. That way we don't give ts users more restrictions than js users.

steffenz: @ivarni That makes a lot of sense, and I agree. Thanks for your feedback. 
@hdaljord That's great to hear, and I agree with having looser ts types in such cases. Thanks for the feedback. 

If you think this is an acceptable solution, it would be great if you could mark it as approved :smile:

hdaljord (approved):

PR #865, opened 2020-03-04
fix(ffe-form-react): updated type for input

hdaljord:
Input now extends React.DetailedHTMLProps to make it compatible with external libraries.

We have started to use react-hook-form and somehow the old definition was not enough for this component, the ref-definition in the React.InputHTMLAttributes<HTMLInputElement> did not accecpt ref={register()} (register from react-hook-form).


hdaljord:
The library is changed now, so I close this PR.

PR #870, opened 2020-03-10
fix(ffe-lists): Change content-value of checklist__item

ssandoy:
Change content-value of checklist__item so that screen readers reads CHECK MARK instead of BULLET for checkmarks and CROSS MARK instead of X BALLOT for crossmarks


steffenz:
Hi! Thanks for your contribution. I wasn't aware that readers even consumed these things from the CSS, but knowing that your change makes a lot of sense. I would love to try it out myself before approving - could you give me a list of the software/setup you use in order to test this? :smile:

ssandoy:
Hi! I just used the VoiceOver-proram on Mac to read the content of the screen :)

steffenz:
Ah, I see! Unfortunately I'm on Linux, but hopefully I can get a hold of either a Mac or Windows machine and check this out. Thanks for your reply, I'll get back to you as soon as I have it tested.

antidecaf (approved):

PR #881, opened 2020-04-09
feat(ffe-cards-react): add typescript types

ivarni:

### DIFF
--- a/packages/ffe-cards-react/package.json
+++ b/packages/ffe-cards-react/package.json
@@ -6,19 +6,22 @@
   "author": "SpareBank 1",
   "files": [
     "lib",
-    "es"
+    "es",
+    "types"
   ],
   "main": "lib/index.js",
   "module": "es/index.js",
+  "types": "types/index.d.ts",
   "sideEffects": false,
   "repository": {
     "type": "git",
     "url": "ssh://git@github.com:SpareBank1/designsystem.git"
   },
   "scripts": {
-    "build": "npm run build:cjs && npm run build:es",
+    "build": "npm run build:cjs && npm run build:es && npm run build:types",
     "build:cjs": "babel -d lib/. --root-mode=upward  --env-name=cjs --ignore=*.spec.js src/.",
     "build:es": "babel -d es/. --root-mode=upward --env-name=es --ignore=*.spec.js src/.",
+    "build:types": "copyfiles -f src/index.d.ts types/",
     "lint": "eslint src/.",
     "test": "jest",
     "test:watch": "jest --watch"
--- a/packages/ffe-cards-react/src/index.d.ts
+++ b/packages/ffe-cards-react/src/index.d.ts
@@ -1,0 +1,49 @@
+import * as React from 'react';
+
+export interface ComponentBaseProps
+    extends React.HtmlHTMLAttributes<HTMLElement> {
+    className?: string;
+    element?: HTMLElement | string | React.ElementType;
+    children?: React.ReactNode;
+}
+
+export interface CardBaseProps extends ComponentBaseProps {
+    to?: string;
+}
+
+export interface IconCardProps extends CardBaseProps {
+    icon: React.ReactNode;
+    condensed?: boolean;
+    greyCharcoal?: boolean;
+}
+
+export interface ImageCardProps extends CardBaseProps {
+    image: React.ReactNode;
+}
+
+export interface TitleProps extends ComponentBaseProps {
+    overflowEllipsis?: boolean;
+}
+
+export interface SubtextProps extends ComponentBaseProps {}
+
+export interface TextProps extends ComponentBaseProps {}
+
+export interface CardNameProps extends ComponentBaseProps {}
+
+export interface TextCardRenderProps {
+    Title: typeof TitleProps;
+    Subtext: typeof SubtextProps;
+    Text: typeof TextProps;
+}
+
+export interface TextCardProps extends CardBaseProps {
+    leftAlign?: boolean;
+    children: React.ReactNode | Function<TextCardRenderProps>;
+}
+
+declare class CardBase extends React.Component<CardBaseProps, any> {}
+declare class IconCard extends React.Component<IconCardProps, any> {}
+declare class ImageCard extends React.Component<ImageCardProps, any> {}
+declare class TextCard extends React.Component<TextCardProps, any> {}
+declare class ComponentBase extends React.Component<ComponentBaseProps, any> {}
--- a/packages/ffe-cards-react/README.md
+++ b/packages/ffe-cards-react/README.md
@@ -10,3 +10,8 @@ npm install --save @sb1/ffe-cards-react
 
 Run Styleguidist from the repository root to see live examples and documentation,
 or see the markdown files next to the component code in `src/`.
+
+## TypeScript definition files
+
+This component supports TypeScript - please update `index.d.ts` if you change any
+of the external methods or properties in this component.


packages/ffe-cards-react/package.json
```diff
@@ -6,10 +6,12 @@
   "author": "SpareBank 1",
   "files": [
     "lib",
-    "es"
+    "es",
+    "types"
   ],
   "main": "lib/index.js",
   "module": "es/index.js",
+  "types": "types/index.d.ts",
```
hdaljord, line 14: You need to add a script for building types and add it to the standard build script. Have a look at another package.json, for instance for ffe-buttons-react.

packages/ffe-cards-react/src/index.d.ts
```diff
@@ -0,0 +1,51 @@
+import * as React from 'react';
+
+export interface CardBaseProps extends React.HTMLAttributes<HTMLDivElement> {
+    children: React.ReactNode;
+    className?: string;
+    element?: React.ReactNode;
```
hdaljord: I'm not quite sure if this is correct. I see that I've used ``element?: HTMLElement | string | React.ElementType;`` for the same propType-definition earlier.

packages/ffe-cards-react/src/index.d.ts
```diff
@@ -0,0 +1,51 @@
+import * as React from 'react';
+
+export interface CardBaseProps extends React.HTMLAttributes<HTMLDivElement> {
```
hdaljord: It's not a given that the resulting element is a div, so you should inherit React.HTMLAttributes<HTMLElement> instead.

packages/ffe-cards-react/src/index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+    condensed?: boolean;
+    greyCharcoal?: boolean;
+}
+
+export interface ImageCardProps extends CardBaseProps {
+    image: React.ReactNode;
+}
+
+export interface ComponentBaseProps
+    extends React.HtmlHTMLAttributes<HTMLParagraphElement> {
```
hdaljord: Use HTMLElement as it's not necessary a paragraph.

packages/ffe-cards-react/src/index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+    className?: string;
+    element?: React.ReactNode;
+    children?: React.ReactNode;
+}
+
+export interface TitleProps extends ComponentBaseProps {
+    overflowEllipsis: boolean;
```
hdaljord: Should not be required

packages/ffe-cards-react/src/index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export interface TextCardRenderProps {
+    Title: typeof TitleProps;
+    Subtext: typeof SubtextProps;
+    Text: typeof TextProps;
+}
+
+export interface TextCardProps extends CardBaseProps {
+    leftAlign?: boolean;
+    children: React.ReactNode | Function<TextCardRenderProps>;
```
hdaljord, line 42: Have you tested this Function-definition?

hdaljord (changes requested):
Perhaps you can add a section to README.md about typescript definitions for this package, as has been done for other packages with ts-defs?

packages/ffe-cards-react/package.json
ivarni: Right, I just copied it into my node_modules folder for testing since npm linking has been broken since we started this monorepo stuff, missed that part.

Question: if all the "build" script does is to copy a file from `src/` to `types/` why not just put it in the `types/` directory in the first place?

packages/ffe-cards-react/src/index.d.ts
ivarni: Right, a node can also be a number or anything else that is renderable but we don't want that here.

packages/ffe-cards-react/src/index.d.ts
ivarni: I wouldn't make a PR with something I haven't tested ;)

packages/ffe-cards-react/package.json
hdaljord: That's a good remark :+1:

Of course, we can put it there. I have no strong opinions on this, I think I just put it in src as all other written files were there, or I noticed a pattern for it somewhere. 

The only argument I see now is that the index.d.ts-files are in src/ in the other packages and it's a plus when the structure is the same across packages.

packages/ffe-cards-react/src/index.d.ts
hdaljord: That's very good! :)

hdaljord (dismissed):

packages/ffe-cards-react/src/index.d.ts
ivarni: It was a major headache actually 😂 
I have 4 hours of experience with TS so far and the first missing declaration I come across is one of the more complex component APIs we have.

packages/ffe-cards-react/src/index.d.ts
ivarni: There's another problem here though. I want my cards to be links so I need to pass inn the `Link` component from react-router (or reach-router) and THAT component again expects a `to` prop which is used in place of `href`.

How do you normally express this in TS? If element is of the `Link` instance from a given library we must be able to pass the `to` prop, but otherwise it makes no sense to do so.

hdaljord: I'm not quite sure if that is possible, at least I have tried to solve that problem. I don't think it is necessary to make it so specific and I think I would just add an optional to-prop as I did for buttons (MinimalBaseButtonProps). 

Maybe some of the ts-users in bank has any suggestions on how to solve this correctly?

packages/ffe-cards-react/src/index.d.ts
hdaljord: Even with a lot more experience, I can imagine :joy:

packages/ffe-cards-react/src/index.d.ts
```diff
@@ -0,0 +1,52 @@
...
+export interface ImageCardProps extends CardBaseProps {
+    image: React.ReactNode;
+}
+
+export interface ComponentBaseProps
+    extends React.HtmlHTMLAttributes<HTMLElement> {
+    className?: string;
+    element?: HTMLElement | string | React.ElementType;
+    children?: React.ReactNode;
+}
```
steffenz:
```suggestion
export interface ComponentBaseProps
    extends React.HtmlHTMLAttributes<HTMLElement> {
    className?: string;
    element?: HTMLElement | string | React.ElementType;
    children?: React.ReactNode;
}
```

packages/ffe-cards-react/src/index.d.ts
```diff
@@ -0,0 +1,52 @@
+import * as React from 'react';
+
+export interface CardBaseProps extends React.HTMLAttributes {
+    children: React.ReactNode;
+    className?: string;
+    element?: React.ReactNode;
+    to?: string;
+}
```
steffenz:
```suggestion
export interface CardBaseProps extends ComponentBaseProps {
    to?: string;
}
```
Trying to use suggestions here - not sure if it will work properly (?). My point here is however, can't you just extend ``CardBaseProps`` in ``ComponentBaseProps``? I see there is a difference in "element", not sure if that's intentional? :thinking: 

Also check out my suggestion for removing ``ComponentBaseProps`` further down in this file.

steffenz:
I've added two suggestions above this comment. Not sure how that works on GitHub, but I've made the change locally and can most certainly push them if you approve. I do however think it should be possible to commit them directly from GitHub through the "suggestion" feature I used.

Pushing directly to somebody else's branch just seems wrong and rude in this case, and there might very well be good reasoning for doing things the way you do (I've just had a quick look at the code). That's why I used a suggestion, because that's what this is. :smiley: 

Also; Is there any reason we want to allow something else besides a few "fixed" HTML elements like a "div" or "span" in the " ``element`` prop? Now everything passes; which potentially allows users to use invalid tags.

hdaljord:
> Also; Is there any reason we want to allow something else besides a few "fixed" HTML elements like a "div" or "span" in the " `element` prop? Now everything passes; which potentially allows users to use invalid tags.

My thoughts on this is that the TS-definition should be as open as the propType-definition, if propTypes allows any element, ts-definition should too.

So, if we want to limit the allowed elements, I think we should do it both for js-users and ts-users, by changing the js-code as well. If we make the ts-definitions too limiting it will only become annoying.

steffenz:
That makes a lot of sense. Thanks for explaining it so clearly, @hdaljord. If everybody is on the same page here, I suggest we implement this as part of the documentation - I'll have a look at it straight away.

packages/ffe-cards-react/src/index.d.ts
ivarni: I've been looking at this for a while but can't spot the difference here.

packages/ffe-cards-react/src/index.d.ts
ivarni: No that's just an oversight, we definetely don't want react-nodes in `element` since that can include numbers and lots of other things.

ivarni:
I don't think you can push to a branch on my fork if you wanted to unless I gave you access. Unless there's something very different with branches that have open PRs.

ivarni:
> I don't think you can push to a branch on my fork if you wanted to unless I gave you access. Unless there's something very different with branches that have open PRs.

I still had to squash it to make all commits follow the commit-rules so not sure how useful the actual merge was but the inline diff-view was nice.

packages/ffe-cards-react/src/index.d.ts
steffenz: I tried to push, but ended up creating a new branch. You can see all the changes to the file there; https://github.com/SpareBank1/designsystem/commit/ab4960626e6fdc40483b0034f2db2ac255336c3a

ivarni: Seems the only effective diff then ends up being the ordering of exports :)

ivarni: I don't have any strong opinions about that so I pushed it

ivarni:
I wouldn't mind getting this merged by the way, my build's been breaking for a while now.

Alternatively I can try to get it into the @types repo or set up som post-install action locally to patch my install of ffe-cards-react.

hdaljord (approved):
Sorry, I thought I'd already approved this one.

PR #884, opened 2020-04-15
fix(ffe-tables-react): dont mutate passed columns

### LINKED ISSUE #883: The table-component is mutating its "columns" prop when passed an "expandableContentMapper"

`ffe-tables-react`

https://github.com/SpareBank1/designsystem/blob/develop/packages/ffe-tables-react/src/Table/Table.js#L28-L36

The `props.columns` array is being modified using push.

I should be able to re-render the component that is using the Table component without getting extra columns.

Every time I re-render my component due to an update from a hook I get another "expand" column.

image

ivarni:
Previously the component was mutating the passed `columns`
prop if an `expandedContentMapper` was being used. This is
not a good pattern to use in React and the component no longer
does this.

Fixes #883

### DIFF
--- a/packages/ffe-tables-react/src/Table/Table.js
+++ b/packages/ffe-tables-react/src/Table/Table.js
@@ -32,12 +32,19 @@ class Table extends Component {
             return null;
         }
 
+        const renderColumns = [...columns];
+
         if (this.props.expandedContentMapper) {
-            columns.push({ key: 'expandIcon', header: '', alignRight: true });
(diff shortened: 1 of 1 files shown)


steffenz (approved):
It makes sense to me to use the "immutable pattern" rather than modifying an existing array. Although I haven't been working extensively with this particular component, I really can't see any problems with the suggested modification. Thanks for creating a separate issue and linking them together (and of course also for your contribution). :smile:

PR #899, opened 2020-06-16
feat(ffe-system-message-react): stop using deprecated lifecycle methods

mortenk:
Do the collapse animation the same way as context-message, instead of using react-collapse/react-motion, which use deprecated lifecycle methods (componentWillReceiveProps)


hdaljord (approved):

hdaljord (changes requested):
Test feiler.

hdaljord (dismissed):

hdaljord (approved):

PR #914, opened 2020-07-30
fix(ffe-form-react): remove checkbox props

ivarni:
BREAKING CHANGE:

Instead of having an `innerRef` prop, use the more modern
approach with `React.forwardRef`. This means that instead
of using the `innerRef` prop, refs should be passed as
they would to any other React element.

```diff
- <Checkbox innerRef={ref} />
+ <Checkbox ref={ref} />
```

While we're doing breaking changes anyway, removed
support from the two deprecated props `label` and
`valid`. Use children and `aria-invalid` instead.

```diff
- <Checkbox invalid={true} label="To arr is pirate" />
+ <Checkbox aria-invalid={true}>To arr is pirate</Checkbox>
```

### DIFF
--- a/packages/ffe-form-react/src/Checkbox.js
+++ b/packages/ffe-form-react/src/Checkbox.js
@@ -1,20 +1,10 @@
 import React, { Fragment } from 'react';
-import { bool, node, string, func, oneOfType, shape, object } from 'prop-types';
+import { bool, node, string, func, oneOfType } from 'prop-types';
 import { v4 as hash } from 'uuid';
 import classNames from 'classnames';
 
-export default function Checkbox(props) {
-    const {
-        children,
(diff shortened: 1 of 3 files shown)


steffenz (changes requested):
Thanks for your contribution :tada: 
However, it seems like your changes broke some of the tests (according to the Travis logs). If you could adjust them to fit your changes, that would be much appreciated :smile:

ivarni:
Sure, I'll take a look. I thought I ran them locally but it's possible I forgot 🦡

ivarni:
Oh, clearly I didn't run them 😂

kwltrs (approved):

steffenz (approved):

PR #1105, opened 2021-05-12
fix(ffe-datepicker-react): make the datepicker work in shadow dom

MarkusSN:
We should always close an opened datepicker, whenever a click happens
outside the datepicker. To enable this, the datepicker registers a
global click handler on window, that checks the origin of every click
event.

Previously, the global click handler used Element.contains() to check if
the event target was contained in the components DOM tree. This does not
work if the component is rendered in a shadow DOM, since events
originating from a shadow DOM are retargeted, and thus the event target
is not contained in the components DOM tree. The result is that the same
click event that opens the datepicker, also closes the datepicker when
the event bubbles to window.

We make the datepicker work in shadow DOM by setting a flag on all click
events originating from the datepicker, and change the logic in the
global click handler to check for this flag instead of using
Element.contains().

### DIFF
--- a/packages/ffe-datepicker-react/src/datepicker/Datepicker.js
+++ b/packages/ffe-datepicker-react/src/datepicker/Datepicker.js
@@ -172,10 +172,7 @@ export default class Datepicker extends Component {
     }
 
     globalClickHandler(evt) {
-        if (
-            this.state.displayDatePicker &&
-            !this._datepickerNode.contains(evt.target)
-        ) {
+        if (this.state.displayDatePicker && !evt.__isEventFromFFEDatepicker) {
             this.closeCalendar();
         }
     }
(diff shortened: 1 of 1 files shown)


MarkusSN:
Jeg har testet denne på react v16 og v17

packages/ffe-datepicker-react/src/datepicker/Datepicker.js
```diff
@@ -301,8 +308,10 @@ export default class Datepicker extends Component {
                             ? `ffe-datepicker-label-${this.datepickerId}`
                             : undefined
                     }
+                    role="none"
```
kwltrs, line 311: Er dette attribut relatert til resten av pr-en?

kwltrs (approved):

packages/ffe-datepicker-react/src/datepicker/Datepicker.js
MarkusSN: Ja, jeg fikk en feil på linting uten role, siden jeg setter en onClick på en div: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md

Usikker på om dette er beste løsning, ser at det kommer en warning slik det er nå

PR #1129, opened 2021-06-17
fix(designsystem): adjust makefile to BSD `find`

kwltrs:
This change fixes a problem with `make` on Mac OS because of Mac OS using BSD find instead of GNU findutils, which is more common on Linux.


HeleneKassandra (approved):
WOoo!!

PR #1138, opened 2021-06-24
fix(ffe-account-selector-react): Add selectedAccount not found in acc…

MariaOverlierBerg:
…ounts list to accounts list

## Beskrivelse

Setter selectedAccount i dropdownList dersom den ikke allerede eksisterer i lista.


fossecode (approved):

packages/ffe-account-selector-react/src/components/account-selector/AccountSelector.js
```diff
@@ -61,7 +73,7 @@ const AccountSelector = ({
     const onInputChange = event => {
         setInputValue(event.target.value);
         if (inputProps?.onChange) {
-            inputProps.onChange();
+            inputProps.onChange(event);
```
fossecode, line 76: Trengs `event` her?

PR #1153, opened 2021-07-12
feat(ffe-searchable-dropdown-react): focus clear button instead of input field on option click

torkjels:
Focusing the input field would make the keyboard pop up on mobile devices
even if the user only clicked the toggle button
and selected straight from the dropdown like a regular select element.


MariaOverlierBerg:
Det var lurt å skille mellom open- og clear-knappene!

MariaOverlierBerg (dismissed):

MariaOverlierBerg (approved):

PR #1565, opened 2023-01-19
fix(ffe-form-react): make text alig right ts type optional

pethel:


HeleneKassandra (approved):
Takk for at du rettet det!

PR #1881, opened 2024-04-15
fix(ffe-webfonts): removed woff from sb1-fonts-inline.less

pethel:
Fullfører den her 

https://github.com/SpareBank1/designsystem/pull/1854


HeleneKassandra:
Kan vi slette de i fonts.css også? Og fjerne filene?

HeleneKassandra (approved):

PR #3160, opened 2026-04-13
fix(ffe-form-react): allow ReactNode for TextField prefix and suffix

alexander-svendsen:
Widen prefix and suffix props from string to React.ReactNode and use Omit to prevent conflicts with the native input element's own prefix and suffix attributes.

The native prefix attribute on `<input>` is a legacy/non-standard attribute. it has no meaningful effect in modern browsers.
It was part of an older XForms spec and is essentially unused today.


antidecaf (approved):

dagfrode (approved):
Takk!

PR #3162, opened 2026-04-15
fix(ffe-datepicker-react): allow key events to bubble from SpinButton input fields

ivarni:
## Summary

- Removes `stopPropagation()` from `SpinButton`'s `handleKeyDown` handler so key events bubble from the input fields consistently with how they behave when the calendar is open
- Keys that need to be consumed for UX purposes (arrow keys) still call `preventDefault()` to prevent scroll/browser defaults

## Test plan

- [ ] Open the Datepicker story in Storybook and verify typing/navigating in the input fields works as expected
- [ ] Verify key events bubble to parent components (e.g. wrapping a datepicker in a keydown listener should now receive events from the input fields)
- [ ] All 99 existing tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dagfrode:
Innlemmet i branch med breaking changes #3147

